PR #2, opened 2017-05-17
Provide assets for color space test.

Hixie:
This is for <https://github.com/flutter/flutter/pull/10000>.

I considered using the Gallery assets repo, but it seems wrong to
include these assets in the gallery app, so I decided to use this repo
instead.


abarth:
LGTM

PR #4, opened 2017-05-20
Add a diagram for AppBar

Hixie:
cc @abarth


abarth:
LGTM

sethladd:
@InMatrix is this useful for widget gallery?

Hixie:
Probably not as-is, but if you want me to produce icons for the widget gallery, assign the bug to me and I'll get to it in due course.

InMatrix:
You mean the widget catalog? I think the current illustration of AppBar borrowed from Material Design is sufficient. This diagram provides useful detail and it is more appropriate in the API doc.

PR #5, opened 2017-05-24
InkWell and InkResponse images

Hixie:


HansMuller:
LGTM

As the structure of the image generating apps stabilizes you may want to document how to use things.

Hixie:
Yeah I'm trying to learn what the right answers are.

PR #7, opened 2017-05-31
Some diagrams for BoxFit.

Hixie:


Hixie:
@HansMuller

painting/box_fit.dart
```diff
@@ -0,0 +1,100 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:ui' as ui;
+
+import 'package:flutter/material.dart';
+
+void main() {
+  List<Widget> children = <Widget>[];
```
HansMuller: Our linter would have crabbed about this not being final :-)

HansMuller:
This is pretty slick. All those diagrams for just a little code. NICE.

LGTM

painting/box_fit.dart
Hixie: la la la? :-)

PR #11, opened 2017-06-19
Making assets repo easier to build.

gspencergoog:
I was going to add some code here to generate my docs, but it seemed to involve a lot of manual/local tweaks just to run it, so before I add my code, I thought I'd rearrange the structure a bit, making a src dir to separate the code from the assets and do "flutter create", and add a generate.sh script that will (mostly) automate generating assets.  The exception is "colors.dart", which requires more manual steps, but I documented those.

I didn't move the original asset directories, since those paths are used for serving assets to the docs, and all the docs would have to be updated.


src/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+### Prerequisites
+
+  - ImageMagick (`brew install imagemagick`).
+  - optipng (`brew install optipng`).
+
+## Generating color swatches
+
+Because there's no way to pass startup parameters to a program (and currently
+--route [doesn't work](https://github.com/flutter/flutter/issues/10813)), the
```
Hixie: this is fixed now

src/animation/curve.dart
```diff
@@ -170,11 +170,12 @@ void main() {
   new Timer(const Duration(seconds: 1), () {
     print('The following commands extract out the ${curves.length} images from a screenshot file.');
     print('You can obtain a screenshot by pressing "s" in the "flutter run" console.');
-    print('export FILE=flutter_01.png # or whatever the file name is');
+    print('Or, use the flutter screenshot command while the app is running.');
```
Hixie: why do you add this here but not to, e.g., tile_mode?

src/generate.sh
```diff
@@ -0,0 +1,87 @@
...
+elif [[ "$1" == "--vertical" ]]; then
+    echo "Be sure the device is in a vertical orientation..."
+    apps=("${vertical_apps[@]}")
+else
+    apps=("$@")
+fi
+
+for app in "${apps[@]}"; do
+    (collect_image "${app}")
+done
```
Hixie: It may be worth building this in dart instead of bash, for better maintainability. Any time a bash script uses an if statement...

src/material/app_bar.dart
```diff
@@ -77,9 +77,9 @@ class _DiagramState extends State<Diagram> {
                       key: flexibleSpace,
                       decoration: new BoxDecoration(
                         gradient: new LinearGradient(
-                          begin: new FractionalOffset(0.50, 0.0),
-                          end: new FractionalOffset(0.48, 1.0),
-                          colors: [Colors.blue.shade500, Colors.blue.shade800]
+                            begin: new FractionalOffset(0.50, 0.0),
+                            end: new FractionalOffset(0.48, 1.0),
+                            colors: [Colors.blue.shade500, Colors.blue.shade800]
```
Hixie: extraneous additional indent

src/material/lib/utils.dart
```diff
@@ -125,11 +125,11 @@ class Labeller extends CustomPainter {
...
-      'The following command extracts the image from a screenshot file.\n'
-      'You can obtain a screenshot by pressing "s" in the "flutter run" console.\n'
-      'Make sure the whole diagram is visible (you may need to rotate the device).\n'
-      'export FILE=flutter_01.png # or whatever the file name is\n'
-      'convert \$FILE -crop ${w}x$h+$x+$y -resize \'450x450>\' $filename'
+        'The following command extracts the image from a screenshot file.\n'
+        'You can obtain a screenshot by pressing "s" in the "flutter run" console.\n'
+        'Make sure the whole diagram is visible (you may need to rotate the device).\n'
+        'BASH: export FILE=flutter_01.png # or whatever the file name is\n'
+        'BASH: convert \$FILE -crop ${w}x$h+$x+$y -resize \'450x450>\' $filename'
```
Hixie: extraneous additional indent

Hixie:
Another way to do this would be to use `flutter driver` scripts, now that driver scripts can just communicate directly with the app.

Hixie:
If you're making a dedicated app for generating the screen shots, it may be worth adding a Travis test that runs the analyzer on the scripts (with the Flutter repo's strict analysis options).

src/animation/curve.dart
gspencergoog: No idea.  I think I meant to put it everywhere, and then forgot.

gspencergoog:
I've converted the bash script to a dart script (which was really fun, actually: I've never made command-line dart before).  The color swatches are now generated automatically too.

The bug with not being able to run a new app without doing a clean build really slows things down...

gspencergoog:
> If you're making a dedicated app for generating the screen shots, it may be worth adding a Travis test that runs the analyzer on the scripts (with the Flutter repo's strict analysis options).

Is there documentation on how to add a Travis test?

Hixie:
> The bug with not being able to run a new app without doing a clean build really slows things down...

Embrace the yak shave? :-)

> Is there documentation on how to add a Travis test?

Docs: https://docs.travis-ci.com/

https://github.com/flutter/flutter/blob/master/.travis.yml is how we've set it up for Flutter's main repo.
https://github.com/flutter/engine/blob/master/.travis.yml is how we set up the engine repo.

I've flipped the switch on Travis itself so that it'll run scripts for this repo.

gspencergoog:
> Embrace the yak shave? :-)

Hey, only one yak at a time, please! I did look into it, but I'll wait to "lick the cookie" until after this is in: I do have a day job, after all. :-)

Hixie:
Can you try to see if the problem you were running into with `flutter run foo.dart; flutter run bar.dart` is gone? I suspect it may have been caused by the fix for https://github.com/flutter/flutter/issues/7673 (which just got reverted).

gspencergoog:
OK, so I've been looking at activating Travis for this repo, and it looks to me like the easiest thing to do is to make it look like the plugins repo: use a slightly modified version of the .travis.yaml there, and the [plugin tools](https://github.com/flutter/plugin_tools) app to run the stages.  Unfortunately, the plugin tools are (not surprisingly) specific to the organization of the plugins repo.  I could just rearrange this repo so that each app that draws a diagram is a separate "package" in a "packages" directory, and then use the plugin tools as-is, but that requires a lot of extra boilerplate in this repo, or I could try and modify the plugin_tools to have some extra arguments to make it more general.  Which one do you think makes more sense?

Hixie:
There's only two things that Travis needs to do: run the analyzer over all the code (`flutter analyze` in the app directory) and run the tests, once we have some (`flutter test` in the app directory). I don't think you need anything more fancy than that.

gspencergoog:
OK, that's significantly less work if I don't need to check format or build.

gspencergoog:
> Can you try to see if the problem you were running into with flutter run foo.dart; flutter run bar.dart is gone? I suspect it may have been caused by the fix for flutter/flutter#7673 (which just got reverted).

No, the problem is still there.

Hixie:
How about now?

gspencergoog:
Yes!  Now it is fixed.

src/android/app/src/main/java/com/yourcompany/src/MainActivity.java
```diff
@@ -0,0 +1,14 @@
+package com.yourcompany.src;
```
Hixie: let's not use "com.yourcompany"

src/generate.dart
```diff
@@ -0,0 +1,283 @@
...
+
+/// The type of the process runner callback.  This allows us to
+/// inject a fake process runner into the DiagramGenerator for tests.
+typedef Future<ProcessResult> ProcessRunner(String executable, List<String> arguments,
+    {String workingDirectory,
+    Map<String, String> environment,
+    bool includeParentEnvironment,
+    bool runInShell,
+    Encoding stdoutEncoding,
+    Encoding stderrEncoding});
```
Hixie: i think if you add a trailing comma here you may get slightly cleaner results from dartfmt. The above is pretty terrible.

Hixie: ditto various others below

README.md
```diff
@@ -37,4 +37,4 @@ than by hand. For example, have a little Flutter program that
 generates the PNGs so that we can easily re-run the program and
 regenerate the images if we want to e.g. change the resolution or
 adjust the text slightly. For an example,
-see [tile_mode.dart](dart-ui/tile_mode.dart).
+see [tile_mode.dart](src/dart-ui/tile_mode.dart).
```
Hixie: you should update this with instructions on how to do it automatically

Hixie: as in, how to use the script you wrote

Hixie:
LGTM, but travis is angry.

Hixie:
Does running the script result in any changes to the PNGs?

README.md
gspencergoog: Actually, I did write a README.md with that info, I just put it in the src subdir.  I'll move it here.

src/android/app/src/main/java/com/yourcompany/src/MainActivity.java
gspencergoog: Done.  I used io.flutter instead.

src/generate.dart
gspencergoog: I did a pass fixing up formatting now that I know a bit more about formatting Flutter code.

gspencergoog:
Yes, it does, but I've been running on a Nexus 5x which doesn't have a big enough screen to capture them all.  I can do a run on a Pixel XL and see what changes if you like.

PR #12, opened 2017-11-03
Updating all the images with the new generate script.

gspencergoog:
There are some differences from the originals, but hopefully these represent changes/bug fixes since the last time they were generated.


Hixie:
LGTM

Hixie:
The card.png one has a broken icon now for some reason (and is really tiny and ugly).

PR #16, opened 2018-01-08
Spring cleaning

Hixie:
cc @gspencergoog for review


src/generate.dart
```diff
@@ -74,90 +85,133 @@ class DiagramGenerator {
...
-      print("Failed to run command '$flutterCommand screenshot --out=$outputName' in ${tmpDir.path}:\n${processResult.stderr}");
-      return false;
+  static const Duration timeout = const Duration(seconds: 60);
+
+  Future<List<String>> runFlutter({
+    @required List<String> args,
+    @required String until,
+    AsyncCallback then,
+  }) async {
+    final Process process = await processStarter(flutterCommand, args, workingDirectory: projectDir);
```
gspencergoog: This could use the ProcessManager instead, from [package:process](https://pub.dartlang.org/packages/process).  This allows us to mock the ProcessManager instead of needing to mock the start and run methods and have the typedefs, etc.

src/generate.dart
```diff
@@ -74,90 +85,133 @@ class DiagramGenerator {
...
+    final List<String> lines = <String>[];
+    process.stdout.transform(UTF8.decoder).transform(const LineSplitter()).listen((String data) {
+      lines.add(data);
+    });
+    process.stderr.transform(UTF8.decoder).transform(const LineSplitter()).listen((String data) {
+      lines.add(data);
```
gspencergoog: Don't you want to separate stderr output from the stdout output, and only look on one for the "until" string?  Otherwise you might get interleaved blocks in the output.

src/generate.dart
```diff
@@ -74,90 +85,133 @@ class DiagramGenerator {
...
+    });
+    final DateTime startWait = new DateTime.now();
+    while (!lines.any((String line) => line.contains(until))) {
+      if (new DateTime.now().difference(startWait) > timeout) {
+        stdout.write('\n');
+        print('Timed out.');
+        print('Output from application:');
+        lines.forEach(print);
```
gspencergoog: Wouldn't you want to write stdout to stdout (via print), and stderr to stderr?

src/generate.dart
```diff
@@ -166,118 +220,98 @@ class DiagramGenerator {
...
   Future<Null> _optimizeImages() async {
-    Directory destDir = new Directory(path.joinAll(path.split(projectDir)
+    final Directory destDir = new Directory(path.joinAll(path.split(projectDir)
       ..removeLast()
       ..add(_diagramType)));
-    List<String> images = (await tmpDir.list().toList()).map((FileSystemEntity e) => e.path).toList();
+    final List<String> images = (await temporaryDirectory.list().toList()).map((FileSystemEntity e) => e.path).toList();
     for (String imagePath in images) {
-      FileSystemEntityType type = await FileSystemEntity.type(imagePath);
+      final FileSystemEntityType type = FileSystemEntity.typeSync(imagePath);
```
gspencergoog: Since this function has to be async anyhow for the directory listing, what is the advantage of using typeSync (and the other *Sync functions below)?

src/generate.dart
Hixie: Unless you are particularly concerned about this, I suggest we leave it as-is, since this code will all end up going by the wayside when we move to the `flutter drive` approach we discussed earlier today.

src/generate.dart
Hixie: Only complete lines get added, so interleaving shouldn't really matter (and will help when the output is dumped to the console in case of error).

src/generate.dart
Hixie: What would the functional difference be? I mean it all just ends up on the console, right.

src/generate.dart
Hixie: See https://github.com/flutter/flutter/issues/7166 and https://github.com/dart-lang/linter/commit/42e227ad7268a569418349bc4cbbd04dcec220 -- the linter complains about these specific methods.

src/generate.dart
gspencergoog: OK, works for me.

src/generate.dart
gspencergoog: If someone redirects output, it'll go different places.  It's not hugely important in this case.

src/generate.dart
gspencergoog: OK

gspencergoog:

PR #18, opened 2018-01-11
Tell Jekyll not to expose the src directory.

Hixie:
cc @gspencergoog 

Right now Jekyll isn't building the assets site because the src/ directory now has a symlink that points somewhere outside the repo. This is intended to make it ignore the src/ directory.


gspencergoog:

PR #20, opened 2018-04-10
Generate sweep gradient assets

xqwzts:
This PR modifies the tile_mode generator to also generate screenshots for Sweep gradients (ref: https://github.com/flutter/engine/pull/4950).

To keep all the tile_mode screenshot generation in one library, I had to change from the previous script that layed out both linear and radial gradients in the same a row [with three rows in total, one for each tile mode] to one that follows the [blend_mode](https://github.com/flutter/assets-for-api-docs/blob/master/src/dart-ui/blend_mode.dart) generator's pattern of having an individual mode per page.

That change allows for more space [width] for each item, so the new screenshots are double the width of the previous ones [400x200 vs 200x200].

PR #21, opened 2018-04-10
Add sweep gradient assets

xqwzts:
Adds screenshots for the new sweep gradients (ref: https://github.com/flutter/engine/pull/4950).

Updates the existing tile_mode screenshots, changing their dimensions.

The generator used and more details can be found here: https://github.com/flutter/assets-for-api-docs/pull/20.


gspencergoog:
Did you add any new code to generate these?  If so, could you add that to this PR so that we can regenerate them in the future?

xqwzts:
Hey @gspencergoog - the PR at #20 includes the code that generates all of these, but it can't land until the `SweepGradient` widget it uses lands in flutter [which I haven't pushed a PR for yet] - and `SweepGradient` can't land until `Gradient.sweep` lands in the engine - which includes these screenshots in its docs.

So it's a bit of a dependency mess, but I was hoping we could add the screenshots now, add the engine changes with the screenshots in the docs, then add the flutter changes once it rolls to the new engine - which will allow us to land the generator code here :)

gspencergoog:
No, I think we should leave them out of the docs until the functionality is available on the framework side too. Nobody's able to use the sweep gradient until it's in the framework anyhow.

You can then check in the new images in a PR along with the generator code at the same time.

Hixie:
I was actually going to suggest this order:
- land the code for the new assets (which won't work for now), with the new screenshots
- land the engine patch
- roll the engine (which updates the framework docs)

It's not a big deal if the assets-for-api-docs repo has extraneous images or if the code here doesn't fully compile yet without special patches, so long as that is a temporary situation.

xqwzts:
I'm closing this PR since the consensus is to land the screenshots with the generator script changes - which are available together in #20 . Since I've done some updates there these screenshots are now obsolete. Thanks :)

littlen101:
Hello wondering is there an estimate as to when the SweepGradient will be available to the framework

gspencergoog:
It is [already available](https://github.com/flutter/flutter/blob/35346b4966c1558b1164d3b2f5fd5158d04c41ff/packages/flutter/lib/src/painting/gradient.dart#L763). It has not yet made it to the beta channel, but will with our next beta roll.

littlen101:
Thank you for the update

PR #23, opened 2018-04-23
Add a Diagram package for making diagrams.

gspencergoog:
This uses the newly added toImage capability for RenderRepaintBoundary to draw diagrams and capture images.

I've converted one of the old diagrams ("curves") into the new framework as a proof of concept.  More will follow in other PRs.

Much of this PR is boilerplate for the package.  The interesting bits are:
- `packages/diagram/lib/diagram.dart`
- `packages/diagram/test/diagram_test.dart`
- `packages/diagram/example/simple/lib/main.dart`
- `diagrams/lib/curve.dart`
- `diagrams/lib/diagram_step.dart`
- `diagrams/lib/main.dart`


packages/diagram/example/simple/.metadata
```diff
@@ -0,0 +1,8 @@
+# This file tracks properties of this Flutter project.
+# Used by Flutter tool to assess capabilities and perform upgrades etc.
+#
+# This file should be version controlled and should not be manually edited.
+
+version:
+  revision: fb86332cd1580ee0a4e02b1fdcc629371caad499
+  channel: widget_image
```
Hixie: heh

gspencergoog: Heh. Whoops.  I guess it got it from my currently open flutter repo.

packages/diagram/example/simple/lib/main.dart
```diff
@@ -0,0 +1,64 @@
...
+// found in the LICENSE file.
+
+import 'package:diagram/diagram.dart';
+import 'dart:async';
+import 'dart:io';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:flutter/widgets.dart';
+import 'package:flutter/material.dart';
+import 'package:path_provider/path_provider.dart';
+import 'package:path/path.dart' as path;
```
Hixie: nit: import order

packages/diagram/lib/diagram.dart
```diff
@@ -0,0 +1,442 @@
+// Copyright 2018 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+library diagram;
```
Hixie: we generally don't include the library line unless we're documenting the library itself

packages/diagram/lib/diagram.dart
```diff
@@ -0,0 +1,442 @@
...
+import 'package:flutter/rendering.dart';
+import 'package:flutter/scheduler.dart';
+import 'package:flutter/services.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:path/path.dart' as path;
+import 'package:vector_math/vector_math_64.dart';
+
+// The diagram host widget. Diagrams are wrapped by this widget to provide
+// the needed structure for capturing them.
+class _Diagram extends StatefulWidget {
```
Hixie: looks like this could be stateless

packages/diagram/lib/diagram.dart
```diff
@@ -0,0 +1,442 @@
...
+  ///
+  /// This is essentially the same as [toMatrix] but ignoring the device pixel
+  /// ratio.
+  ///
+  /// This is useful because pointers are described in logical pixels, as
+  /// opposed to graphics which are expressed in physical pixels.
+  Matrix4 toHitTestMatrix() => _hitTestMatrix.clone();
+
+  @override
+  String toString() => 'TestViewConfiguration';
```
Hixie: copypasta

packages/diagram/lib/diagram.dart
```diff
@@ -0,0 +1,442 @@
...
+
+  /// Provides the transformation matrix that converts coordinates in the test
+  /// coordinate space to coordinates in logical pixels on the real display.
+  ///
+  /// This is essentially the same as [toMatrix] but ignoring the device pixel
+  /// ratio.
+  ///
+  /// This is useful because pointers are described in logical pixels, as
+  /// opposed to graphics which are expressed in physical pixels.
+  Matrix4 toHitTestMatrix() => _hitTestMatrix.clone();
```
Hixie: i'm guessing you don't actually need toHitTestMatrix ?

packages/diagram/lib/diagram.dart
```diff
@@ -0,0 +1,442 @@
...
+
+  @override
+  String toString() => 'TestViewConfiguration';
+}
+
+// Provides a binding different from the regular Flutter binding so that
+// diagrams can control their timeline and physical device size.
+class _DiagramFlutterBinding extends BindingBase
+    with
+        GestureBinding, //
```
Hixie: stray //

packages/diagram/lib/diagram.dart
gspencergoog: So.. I'm not sure what you mean here.  This is the library itself, as far as I'm aware.

packages/diagram/lib/diagram.dart
```diff
@@ -0,0 +1,442 @@
...
+        SchedulerBinding,
+        PaintingBinding,
+        RendererBinding,
+        WidgetsBinding {
+  @override
+  void initInstances() {
+    super.initInstances();
+    // Short circuit the begin frame to use our timestamp instead of the "real" one.
+    // Re-use the previous onBeginFrame so that all of the hot reload, etc. works.
+    dynamic oldBeginFrame = ui.window.onBeginFrame;
```
Hixie: This would be more idiomatic if it was done by overriding handleBeginFrame, if that works.

packages/diagram/lib/diagram.dart
gspencergoog: Well, probably not, but I figured I'd include it in case it's needed to simulate a tap (I'm thinking of your diagram that shows the ripple animation).

packages/diagram/lib/diagram.dart
```diff
@@ -0,0 +1,442 @@
...
+  Duration _timestamp = Duration.zero;
+  GlobalKey _boundaryKey = new GlobalKey();
+
+  /// Determines the ratio between physical units and logical units.
+  ///
+  /// The [pixelRatio] describes the scale between the logical pixels and the
+  /// size of the output image. It is independent of the
+  /// [window.devicePixelRatio] for the device, so specifying 1.0 (the default)
+  /// will give you a 1:1 mapping between logical pixels and the output pixels
+  /// in the image. The output image is not anti-aliased, so you may wish to
```
Hixie: Why is the output image not antialiased?

packages/diagram/lib/diagram.dart
```diff
@@ -0,0 +1,442 @@
...
+  }
+
+  /// Returns the singleton for the binding, after making sure it is
+  /// initialized.
+  static WidgetsBinding instance() {
+    if (WidgetsBinding.instance == null) {
+      new _DiagramFlutterBinding();
+    }
+    return WidgetsBinding.instance;
+  }
```
Hixie: see https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/scheduler/binding.dart#L194 for the idiomatic way to do this.

packages/diagram/lib/diagram.dart
```diff
@@ -0,0 +1,442 @@
...
+    return new _DiagramViewConfiguration(
+      pixelRatio: pixelRatio,
+      size: screenDimensions,
+    );
+  }
+
+  /// Captures an image of the [RepaintBoundary] with the given key.
+  Future<ui.Image> takeSnapshot() async {
+    RenderRepaintBoundary object = _boundaryKey.currentContext.findRenderObject();
+    return await object.toImage(pixelRatio: pixelRatio);
```
Hixie: you can remove the `async` and `await` keywords here with no change to the semantics.

packages/diagram/lib/diagram.dart
```diff
@@ -0,0 +1,442 @@
...
+  /// Advances time by the given duration, and generates a new frame.
+  ///
+  /// The [duration] must not be null, or less than [Duration.zero].
+  void pump({Duration duration: Duration.zero}) {
+    assert(duration != null);
+    assert(duration >= Duration.zero);
+    _timestamp += duration;
+
+    handleBeginFrame(_timestamp);
+    handleDrawFrame();
```
Hixie: if you're driving the framework directly, do you even need to override the onBeginFrame? I'm confused as to your intent here.

packages/diagram/lib/diagram.dart
```diff
@@ -0,0 +1,442 @@
...
+    AnimationFrameFilenameGenerator frameFilenameGenerator,
+    double pixelRatio: 1.0,
+    Size screenDimensions: _kDefaultDiagramViewportSize,
+  }) : assert(builder != null) {
+    this.frameFilenameGenerator = frameFilenameGenerator;
+    outputDirectory ??= Directory.current;
+    _binding.pixelRatio = pixelRatio;
+    _binding.screenDimensions = screenDimensions;
+    this.builder = builder;
+  }
```
Hixie: The idiomatic thing would be to construct the binding here and assert that it's the right type

Hixie: see e.g. https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/widgets/binding.dart#L702
(but with an additional assert)

packages/diagram/lib/diagram.dart
```diff
@@ -0,0 +1,442 @@
...
+
+  /// The generator for filenames when calling drawAnimatedDiagramToFile.
+  ///
+  /// If the returned filenames are relative paths, they will be relative to
+  /// [outputDirectory].
+  AnimationFrameFilenameGenerator get frameFilenameGenerator => _frameFilenameGenerator;
+  AnimationFrameFilenameGenerator _frameFilenameGenerator = _basicFrameFilenameGenerator;
+  set frameFilenameGenerator(AnimationFrameFilenameGenerator frameFilenameGenerator) {
+    _frameFilenameGenerator = frameFilenameGenerator ?? _basicFrameFilenameGenerator;
+  }
```
Hixie: this setter violates the rule that `a = b` should mean that `a == b` for all values of b, when `b == null`.

packages/diagram/lib/diagram.dart
```diff
@@ -0,0 +1,442 @@
...
+    ui.ImageByteFormat format: ui.ImageByteFormat.png,
+  }) async {
+    assert(outputFile != null);
+    if (!outputFile.isAbsolute && outputDirectory != null) {
+      // If output path is relative, make it relative to the output directory.
+      outputFile = new File(path.join(outputDirectory.absolute.path, outputFile.path));
+    }
+    ui.Image captured = await drawDiagramToImage();
+    ByteData encoded = await captured.toByteData(format: format);
+    List<int> bytes = encoded.buffer.asUint8List().toList();
```
Hixie: missing final?

packages/diagram/lib/diagram.dart
```diff
@@ -0,0 +1,442 @@
...
+    print('Writing ${bytes.length} bytes, ${captured.width}x${captured.height} '
+        '${_byteFormatToString(format)}, to ${outputFile.absolute.path}.');
```
Hixie: to avoid the problem of the trailing `.` looking like it's part of the filename, you could remove it and instead put a colon after `to`.

packages/diagram/lib/diagram.dart
```diff
@@ -0,0 +1,442 @@
...
+    return outputImages;
+  }
+
+  /// Draws multiple frames of the widget generated by [builder] into multiple
+  /// files.
+  ///
+  /// The filenames are determined by [frameFilenameGenerator].
+  ///
+  /// The files will be written in whatever format the suffix on the file
+  /// indicates. Valid suffixes are ".jpg", ".jpeg", ".png", ".gif", and ".tga".
```
Hixie: I'm sure glad we still support Truevision TGA in 2018.

packages/diagram/lib/diagram.dart
```diff
@@ -0,0 +1,442 @@
...
+    Duration now = start;
+    int index = 0;
+    List<File> outputFiles = <File>[];
+    while (now <= end) {
+      File outputFile = _getFrameFilename(now, index);
+      ui.Image captured = await drawDiagramToImage();
+      ByteData encoded = await captured.toByteData(format: format);
+      List<int> bytes = encoded.buffer.asUint8List().toList();
+      print('Writing frame $index ($now), ${bytes.length} bytes, ${captured.width}x${captured.height} '
+          '${_byteFormatToString(format)}, to ${outputFile.absolute.path}');
```
Hixie: well here you're just missing the `.`. Maybe add the colon after `to` for grammar pedancy purposes?

packages/diagram/lib/diagram.dart
```diff
@@ -0,0 +1,442 @@
...
+    return null;
+  }
+
+  File _getFrameFilename(Duration timestamp, int index) {
+    File outputFile = frameFilenameGenerator(timestamp, index);
+    if (!outputFile.isAbsolute && outputDirectory != null) {
+      // If output path is relative, make it relative to the output directory.
+      outputFile = new File(path.join(outputDirectory.absolute.path, outputFile.path));
+    }
+    assert(outputFile.path.endsWith('.png'), 'Since the only supported output format is PNG, filenames must end in ".png"');
```
Hixie: this seems to contradict the earlier comment?

packages/diagram/test/diagram_test.dart
```diff
@@ -0,0 +1,173 @@
...
+    constraints: new BoxConstraints.tight(new Size(100.0, 50.0)),
+    child: new Text('Diagram'),
+  );
+}
+
+File filenameGenerator(Duration timestamp, int index) {
+  return new File('test_name_${timestamp.inMilliseconds}_$index.png');
+}
+
+class TestAnimatedDiagram extends StatefulWidget {
```
Hixie: this one also seems to actually be stateless

Hixie:
Mostly LGTM modulo comments.

The name "diagram" might be confusing in the context of code like https://github.com/flutter/assets-for-api-docs/blob/4ff49139bf70eeb233498ca9e5024987dc0333a4/src/material/app_bar.dart#L20 but we could rename that in the transition.

packages/diagram/lib/diagram.dart
Hixie: nothing seems to use it.

packages/diagram/lib/diagram.dart
Hixie: Right but you don't have any dartdocs for the `library`. So you can just omit the `library` line.

packages/diagram/lib/diagram.dart
gspencergoog: Good point. Removed.

packages/diagram/lib/diagram.dart
gspencergoog: LOL.  Yeah, me too.  It *is* one of the simplest formats ever to implement, however.

packages/diagram/lib/diagram.dart
gspencergoog: Yeah, stale comment. Fixed.

packages/diagram/test/diagram_test.dart
gspencergoog: Fixed.

packages/diagram/lib/diagram.dart
gspencergoog: Ahh, so it's useless for anything other than attaching docs to?  I didn't know that.  Removed.

PR #24, opened 2018-04-24
Converted all existing diagrams to use the new framework.

gspencergoog:
This converts the existing diagram code over to using the new framework for capturing diagrams.

Some of the output varies a little bit, because of differences in the pixel ratio and timing.  The variations are minor, and most of the text and drawing looks a lot sharper than the old images.

Not to mention that what once took about 30 minutes and two passes, now takes 7 seconds.

The next step will be to rewrite the generation script to run the app, download the images, run optipng on them all, and save them to the right locations.


analysis_options.yaml
```diff
@@ -0,0 +1,159 @@
...
+#
+# For a list of lints, see: http://dart-lang.github.io/linter/lints/
+# See the configuration guide for more
+# https://github.com/dart-lang/sdk/tree/master/pkg/analyzer#configuring-the-analyzer
+#
+# There are four similar analysis options files in the flutter repos:
+#   - analysis_options.yaml (this file)
+#   - analysis_options_repo.yaml
+#   - packages/flutter/lib/analysis_options_user.yaml
+#   - https://github.com/flutter/plugins/blob/master/analysis_options.yaml
```
Hixie: Five now I guess?

Hixie: oh hey there's a way to import one from the other now. Maybe we can do that here! @devoncarew or @pq may know more about the syntax for doing this.

diagrams/lib/ink_well.dart
```diff
@@ -0,0 +1,105 @@
...
+    final RenderBox target = splashKey.currentContext.findRenderObject();
+    final Offset targetOffset = target.localToGlobal(target.size.bottomRight(Offset.zero));
+    final TestGesture gesture = await controller.startGesture(targetOffset);
+    final List<File> result = <File>[
+      await controller.drawDiagramToFile(
+        new File('ink_well.png'),
+        duration: const Duration(milliseconds: 550),
+      ),
+    ];
+    gesture.up();
```
Hixie: oh this is so much nicer

packages/diagram/lib/diagram.dart
```diff
@@ -244,6 +255,9 @@ class DiagramController {
   /// files.
   Directory outputDirectory;
 
+  double get pixelRation => _binding.pixelRatio;
```
Hixie: ration?

packages/diagram/pubspec.yaml
```diff
@@ -14,5 +14,9 @@ dev_dependencies:
     sdk: flutter
   image: ^1.0.30
 
+test_dependencies:
+  diagram:
+    path: .
```
Hixie: ?

Hixie:
LGTM

analysis_options.yaml
gspencergoog: Yes, that would be great.  Can I do it this way for now, and just switch it in another PR once I know more about how to do it?

diagrams/lib/ink_well.dart
gspencergoog: Agreed.

packages/diagram/lib/diagram.dart
gspencergoog: Whoops!  Clearly not actually used yet... Fixed.

packages/diagram/pubspec.yaml
gspencergoog: Without that, the test couldn't find the diagram package.  I tried `import '../lib/diagram.dart'`, and although that worked, it gave an analyzer warning about not importing things from a 'lib' directory.

analysis_options.yaml
Hixie: Sure.

packages/diagram/pubspec.yaml
Hixie: I'm confused. Where did importing what from where fail?

gspencergoog: The file `package/diagram/test/diagram_test.dart` was importing with `import '../lib/diagram.dart'`, and it didn't fail, it imported properly, but the analyzer did give a warning.

Adding this test dependency made it so that I could do `import 'packages:diagram/diagram.dart'` in the test instead.

analysis_options.yaml
devoncarew: Yeah, you can reference a parent analysis options file (including by a package: url if the package is available in your deps), get all the settings from that file, and choose to override individual settings.

There's an example of that here: https://github.com/flutter/flutter/pull/16979/files#diff-60bf9684f6753f1cf7159920ecc36a89

Thinking about it now, I'd need tp confirm the `package:` reference.

packages/diagram/pubspec.yaml
Hixie: Oh, it's because you have the wrong name in the `name` line in the pubspec.yaml.

PR #25, opened 2018-04-25
Remove old generators, add new generator script.

gspencergoog:
This moves this repo completely over to using the new diagram framework, and allows generation via the `diagrams/generate.dart` script.

Note that this PR currently also include the contents of https://github.com/flutter/assets-for-api-docs/pull/24, which will be merged before this one (and so will not be part of this PR when merged).


Hixie:
Can you land #24 and rebase this one?

gspencergoog:
Yes, that was my plan.

gspencergoog:
OK, All rebased.

README.md
```diff
@@ -32,12 +32,13 @@ the `tests` directory should not be optimised.
 
 ## Generation
 
-See the [documentation for the generate.dart script in the src
-directory](src/README.md), which will will generate a number of
+See the [documentation for the generate.dart script in the `diagrams`
+directory](diagrams/README.md), which will will generate a number of
```
Hixie: you're inconsistent about whether this package should be named `diagram` or `diagrams`, which is leading to that issue with having that hack in the `pubspec.yaml`. I recommend auditing everything to make it all consistent, then you can remove the hack.

diagrams/generate.dart
```diff
@@ -0,0 +1,343 @@
+import 'dart:async';
```
Hixie: license

diagrams/lib/app_bar.dart
```diff
@@ -110,6 +110,7 @@ class AppBarDiagramStep extends DiagramStep {
 
   @override
   Future<List<File>> generateDiagrams() async {
+    controller.pixelRatio = 1.0 / 1.2;
```
Hixie: we should probably make this 1.2 a constant (and add a comment explaining what we're doing here).

diagrams/lib/main.dart
```diff
@@ -67,4 +67,5 @@ Future<Null> main(List<String> arguments) async {
     }
   }
   print('Total elapsed time: ${new DateTime.now().difference(start)}');
+  exit(0);
```
Hixie: ?

Hixie:
LGTM

README.md
gspencergoog: Hmm. Actually, I'm not: the package is called "diagram" (in `packages/diagram`) and the location of the diagrams is in the `diagrams` directory, but that is confusing.  Perhaps I'll just rename the diagram package to "diagram_capture".

diagrams/lib/main.dart
gspencergoog: In order for the generator script to continue and download the generated diagrams, the app needs to actually exit, or the "flutter run" that it invokes hangs.

diagrams/generate.dart
gspencergoog: Good call. Added.

diagrams/lib/app_bar.dart
gspencergoog: Done.

README.md
Hixie: Ah, my bad. I was very confused, either way. :-)
Current status seems better.

diagrams/lib/main.dart
Hixie: Interesting. I wonder if this means we can't run this on iOS?

gspencergoog: Well, we can't run it on iOS anyhow because I used adb to transfer the resulting images back to the host anyhow. Is running on iOS important?

Hixie: ah yeah, good point. We should just mention that in the README somewhere.

PR #27, opened 2018-05-08
Adds conversion of diagram animations to mp4, and animated diagrams for StrokeJoin

gspencergoog:
Our first animated diagrams are here! I've added support for creating animated diagrams as mp4 videos.

miter_4_join
(The real diagram looks a lot nicer than this: GitHub doesn't support MPEG4 files, so this is a dithered GIF).


analysis_options.yaml
```diff
@@ -24,7 +24,6 @@
 
 analyzer:
   language:
-    enableStrictCallChecks: true
```
Hixie: ?

gspencergoog: This option is no longer supported, so it has to be removed from the yaml file.

diagrams/generate.dart
```diff
@@ -308,18 +336,121 @@ class DiagramGenerator {
...
+    for (Map<String, dynamic> metadata in metadataList) {
+      final String prefix = '${metadata['category']}/${metadata['name']}';
+      final File destination = new File(path.join(destDir.path, '$prefix.mp4'));
+      if (destination.existsSync()) {
+        destination.deleteSync();
+      }
+      print('Converting ${metadata['name']} animation to mp4.');
+      await processRunner.runProcess(
+        <String>[
+          ffmpegCommand,
```
Hixie: can we document somewhere that you need ffmpeg? what happens if you run this without it?

gspencergoog: Sure. Added it to the README.md list of prerequisites.

If you run it without it, it will throw an exception and the script will fail.

packages/diagram_capture/test/diagram_capture_test.dart
```diff
@@ -158,22 +160,35 @@ void main() {
...
+      expect(outputFile.existsSync(), isTrue);
+      expect(outputFile.lengthSync(), greaterThan(0));
+
+      Map<String, dynamic> _loadMetadata(File metadataFile) {
+        final Map<String, dynamic> metadata = json.decode(metadataFile.readAsStringSync());
+        final String baseDir = path.dirname(metadataFile.absolute.path);
+        final List<File> frameFiles = metadata['frame_files']
+            .map<File>((dynamic name) => new File(path.normalize(path.join(baseDir, name)))).toList();
+        metadata['frame_files'] = frameFiles;
+        return metadata;
```
Hixie: i guess this is fine since it's a hacky little script but fwiw best practice would be to convert out of this JSONesque map and into fully typed structured data as quick as possible.

packages/diagram_capture/test/diagram_capture_test.dart
```diff
@@ -158,22 +160,35 @@ void main() {
...
+      expect(outputFile.path.endsWith('test_name.json'), isTrue);
       int count = 0;
-      for (File file in outputFiles) {
+      expect(outputFile.existsSync(), isTrue);
+      expect(outputFile.lengthSync(), greaterThan(0));
+
+      Map<String, dynamic> _loadMetadata(File metadataFile) {
+        final Map<String, dynamic> metadata = json.decode(metadataFile.readAsStringSync());
+        final String baseDir = path.dirname(metadataFile.absolute.path);
+        final List<File> frameFiles = metadata['frame_files']
```
Hixie: if we do want to go the JSONesque route, let's at least use constants for the key names.

packages/diagram_capture/test/diagram_capture_test.dart
```diff
@@ -158,22 +160,35 @@ void main() {
...
+        filenameGenerator: filenameGenerator,
         pixelRatio: 1.0,
         screenDimensions: const Size(100.0, 100.0),
       );
 
       controller.builder = (BuildContext context) => new TestAnimatedDiagram(key: key, size: 50.0);
-      final List<File> outputFiles = await controller.drawAnimatedDiagramToFiles(
+      final File outputFile = await controller.drawAnimatedDiagramToFiles(
         end: const Duration(milliseconds: 1200),
         frameDuration: const Duration(milliseconds: 200),
```
Hixie: can we do 60Hz animations instead of 5Hz animations?

Hixie: oh never mind this is a test

diagrams/lib/stroke_join.dart
```diff
@@ -0,0 +1,250 @@
...
+      ),
+    ];
+
+    final List<File> outputFiles = <File>[];
+    for (StrokeJoinPainterWidget stroke in strokes) {
+      print('Drawing stroke diagram for ${stroke.join} (${stroke.filename})');
+      controller.builder = (BuildContext context) => stroke;
+      controller.filenameGenerator = () => new File(stroke.filename);
+      outputFiles.add(
+        await controller.drawAnimatedDiagramToFiles(end: const Duration(seconds: 6), frameDuration: const Duration(milliseconds: 200), metadata: <String, dynamic>{
```
Hixie: can we make this 60Hz rather than 5Hz? (i found the real one this time!)

packages/diagram_capture/test/diagram_capture_test.dart
gspencergoog: Yes, it's just a test.  Also, yes, we can do 60hz animations. I just didn't want to produce too much data for the test.

packages/diagram_capture/test/diagram_capture_test.dart
gspencergoog: OK, I'll build a little wrapper class.

diagrams/lib/stroke_join.dart
Hixie: well this one is going to be an MPEG, right? So that's generally rendered streaming.

Hixie:
LGTM

Don't worry about the structured vs JSON stuff if it's not an obvious win.

diagrams/lib/stroke_join.dart
gspencergoog: Actually, after switching it to 60hz lossless encoding, optimized for cell animations, the file size is about 100K, so I don't think we need to worry. :-)

gspencergoog:
It's an obvious win.  Done.

PR #28, opened 2018-05-15
Copy assets into assets dir in preparation for moving all asset URLs

gspencergoog:
Copy assets into assets dir in preparation for moving all asset URLs to point to assets directory.

The next steps will be to rename the asset URLs, let them propagate to the docs website, and then (once they're all pushed), remove the old diagrams.

I didn't modify any of the diagrams, I just copied them.

Also modified the `generate.dart` script to generate into the new assets directory instead of the old directories, and updated README's to point to the new location.


Hixie:
RSLGTM

PR #29, opened 2018-05-17
Copying some test assets that I forgot

gspencergoog:
In my previous PR, I missed these manual test assets which are referenced by some test documentation.


Hixie:
RSLGTM

PR #30, opened 2018-05-22
Change video encoding parameters to work on Safari/Quicktime

gspencergoog:
I had changed the `ffmpeg` parameters to use "lossless" encoding, but it seems that although Chrome supports that just fine, Safari doesn't have that profile. This PR modifies the encoding parameters for `ffmpeg` to make it "almost lossless", which Safari is fine with, and re-encodes the videos with those parameters.

Also, the interface for the "typed" test function seems to have changed, so I updated the test to reflect that.


Hixie:
RSLGTM

PR #32, opened 2018-05-25
Allows generation of just a subset of the diagrams by passing command line options.

gspencergoog:
Passing either --category or --name to generate.dart will restrict the generated diagrams to just those diagrams that match.  The flags can be specified multiple times to specify more than one.  The category is the directory name that the assets end up in, and the name can be the base name of the diagram, with or without the suffix.


Hixie:
Did you intend the formatting changes to be included? (It's fine if you did, happy to review them, but there have been past PRs where you indicate that was accidental so I thought I'd check before reviewing.)

gspencergoog:
Thanks, no, I didn't intend all of the formatting changes: I had changed my line length for dartfmt to 80 chars while I was working on the dartdoc change, and apparently that isn't a project-specific setting.  I've reverted those.

diagrams/lib/main.dart
```diff
@@ -26,7 +27,8 @@ import 'tile_mode.dart';
 
 Future<Directory> prepareOutputDirectory() async {
   final Directory directory = new Directory(
-    path.join((await getApplicationDocumentsDirectory()).absolute.path, 'diagrams'),
+    path.join(
+        (await getApplicationDocumentsDirectory()).absolute.path, 'diagrams'),
```
Hixie: missing trailing comma for better styling

diagrams/lib/main.dart
```diff
@@ -35,7 +37,20 @@ Future<Directory> prepareOutputDirectory() async {
...
 }
 
-Future<Null> main(List<String> arguments) async {
+Future<Null> main() async {
+  final List<String> arguments = window.defaultRouteName.length > 5
+      ? Uri.decodeComponent(window.defaultRouteName.substring(5)).split(' ')
+      : <String>[];
+  final ArgParser parser = new ArgParser();
+  parser.addMultiOption('category',
+      help: 'Only create the diagrams for the given category.');
```
Hixie: not sure we really need the help lines here, they're redundant with the "real" ones in the main file

diagrams/lib/tile_mode.dart
```diff
@@ -129,8 +129,11 @@ class TileModeDiagramStep extends DiagramStep {
 
     final List<File> outputFiles = <File>[];
     for (TileModeDiagram tileModeDiagram in tileModeDiagrams) {
+      if (onlyGenerate != null && !onlyGenerate.contains(tileModeDiagram.name)) {
+        continue;
+      }
```
Hixie: it's a pity we have to duplicate this pattern everywhere. not sure how we can avoid it... maybe just filter by `category` instead, rather than by any specific filename? That way you could factor the check out into the superclass...

Hixie:
LGTM

diagrams/lib/main.dart
gspencergoog: Fixed.

diagrams/lib/main.dart
gspencergoog: Yeah, I guess so: you can't really see them anyhow.

diagrams/lib/tile_mode.dart
gspencergoog: There already is the ability to filter by category (see `--category` on generate.dart). This just refines it (or provides a separate filter) for filenames.  I thought about making a separate base class that represents a single diagram and has a name (maybe Diagram?).  Each DiagramStep can then provide a list of Diagrams from an accessor, and the base generateDiagrams would call generateDiagram for each one where the filename matches, instead of subclasses overriding generateDiagrams.  I'll see if that works well.

gspencergoog: OK, I refactored it so that the filtering happens in the step.  It does seem to make things a bit cleaner for the implementations.

gspencergoog:
Sorry, I upended things quite a bit when I refactored to move the filtering into the base DiagramStep.  Could you take another look?

I also got rid of the filename generator callback, since it was really not being used, and was redundant with the name parameter to the animation creation routines.

diagrams/generate.dart
```diff
@@ -277,10 +277,18 @@ class DiagramGenerator {
...
+  /// The path to the top of the repo.
   static String get projectDir {
     return path.joinAll(path.split(path.absolute(path.fromUri(Platform.script)))..removeLast());
   }
 
+  /// The output asset directory for all the categories.
+  static String get assetDir {
+    return path.joinAll(path.split(projectDir)
+      ..removeLast()
+      ..add('assets'));
```
Hixie: there's some misleading indentation here. Looks like the cascade applies to the return value of `path.joinAll`, but it's actually applying to the argument.

diagrams/generate.dart
```diff
@@ -469,5 +493,6 @@ Future<Null> main(List<String> arguments) async {
     keepTmp = true;
   }
 
-  new DiagramGenerator(temporaryDirectory: temporaryDirectory, cleanup: !keepTmp)..generateDiagrams();
+  new DiagramGenerator(temporaryDirectory: temporaryDirectory, cleanup: !keepTmp)
```
Hixie: we should rename "keepTmp" to "keepTemporaryDirectory" while we're at it

diagrams/generate.dart
gspencergoog: Fixed.

diagrams/generate.dart
gspencergoog: Done

diagrams/lib/app_bar.dart
```diff
@@ -114,11 +117,12 @@ class AppBarDiagramStep extends DiagramStep {
...
-      await controller.drawDiagramToFile(new File('app_bar.png')),
-    ];
-    return results;
+  Future<List<DiagramMetadata>> get diagrams async => <DiagramMetadata>[const AppBarDiagram(name: 'app_bar')];
+
+  @override
+  Future<File> generateDiagram(DiagramMetadata diagram) async {
+    final AppBarDiagram typedDiagram = diagram;
+    controller.builder = (BuildContext context) => typedDiagram;
+    return await controller.drawDiagramToFile(new File('${diagram.name}.png'));
```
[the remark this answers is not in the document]
gspencergoog: Yeah, I noticed that too.  I thought about maybe just passing the widget to `drawDiagramToFile` and letting the controller make its own builder, but for animated diagrams it seems to make sense to keep the builder pattern. Ultimately, I decided I had done enough refactoring in this PR already. :-)

PR #35, opened 2018-07-17
Update to accomodate change in button size.

gspencergoog:
Fixes a broken test that broke because the button default size changed.


gspencergoog:
cc @Hixie

PR #37, opened 2018-07-18
Curves animation

gspencergoog:
This converts the curve diagrams into animations so that users can see how the curves convert linear time into nonlinear output, with examples of how it can be used.

curve_elastic_in_out

(The "real" animation has a better frame rate: GitHub only supports GIFs).


Hixie:
cc @InMatrix for advice

InMatrix:
Here is an example you might want to take a look: https://easings.net/#easeInOutSine

I thought it looked clean and minimalist (in a good way).

gspencergoog:
OK, I think this is ready for review.  I've run it past @InMatrix and incorporated his feedback.

I've also regenerated all the images, not just the animation curves, to make sure that they're all up to date with the current head of master.

gspencergoog:
Note: this PR currently contains the changes in https://github.com/flutter/assets-for-api-docs/pull/36 as well: I'll merge those out as soon as the other PR lands.

diagrams/lib/curve.dart
```diff
@@ -133,28 +203,237 @@ class CurveDescription extends CustomPainter {
...
+            ),
+          ),
+        ),
+        new Text(
+          name,
+          textAlign: TextAlign.center,
+          style: Theme.of(context).textTheme.body2.copyWith(
+                color: Colors.black,
+                fontSize: 12.0,
+              ),
```
Hixie: trivial nit: weird indentation

Hixie:
LGTM

Hixie:
Looks like Curves.elasticIn misses the last frame.
Also, there's something fishy about the line in ease_out, like it's loosing yarn or something.

PR #38, opened 2018-07-25
Adding animations for transitions.

gspencergoog:
This adds animations for each of the "Transition" widgets that we have (the GIFs here are at a lower frame rate than the "real" ones, and of course on the real page they'll be inside a player that only runs when you click it):

| Transition | Animation |
| ---------- | ---------- |
| [DecoratedBoxTransition](https://docs.flutter.io/flutter/widgets/DecoratedBoxTransition-class.html) | decorated_box_transition |
| [FadeTransition](https://docs.flutter.io/flutter/widgets/FadeTransition-class.html) | fade_transition |
| [PositionedTransition](https://docs.flutter.io/flutter/widgets/PositionedTransition-class.html) | positioned_transition |
| [RelativePositionedTransition](https://docs.flutter.io/flutter/widgets/RelativePositionedTransition-class.html) | relative_positioned_transition |
| [RotationTransition](https://docs.flutter.io/flutter/widgets/RotationTransition-class.html) | rotation_transition |
| [ScaleTransition](https://docs.flutter.io/flutter/widgets/ScaleTransition-class.html) | scale_transition |
| [SizeTransition](https://docs.flutter.io/flutter/widgets/SizeTransition-class.html) | size_transition |
| [SlideTransition](https://docs.flutter.io/flutter/widgets/SlideTransition-class.html) | slide_transition |


Hixie:
Can we factor out the duplicated code?

Also I think the images might be clearer if they were 2 seconds of animation, 1 second pause, 2 seconds animating back, 1 second pause, etc. That would avoid the snapping behaviour we see now.

Hixie:
Also we should probably use the same curve for all of them to make it clearer when you compare them.

gspencergoog:
I'll factor out the duplicated code.  I left it that way because we were going to use the *Example classes for the website as sample code (after some more cleanup/simplification), but the examples can be different code than what generates the diagrams as long as the result is the same.

As for the different curves, the animations won't be appearing on one page together, but separate pages. I chose those because they illustrated the individual transitions better: having an elasticOut for the fade transition just doesn't look all that great, and having just a fastOutSlowIn for the rest seemed too simple. I think I'd rather show a variety of curves to show what is possible instead of making them the same.

I thought about something like it, but didn't originally do the timing you suggest because I haven't implemented a way to trigger gestures on a schedule yet. I have a plan for that though: passing a `Map<Duration, VoidCallback>` to the `drawAnimatedDiagramToFiles` method to trigger callbacks that can make modifications when a particular time is passed. The callbacks can then invoke gestures, modify state, or whatever.

Hixie:
I guarantee that within weeks of you checking these in, someone somewhere will put them all on one page.

I agree that some of them look better with certain curves, though. Maybe have each one have a little caption that says what transition it shows and what curve it's using?
You could even have a sparkline underneath showing the value over time with an animated dot showing where on the graph you are...

packages/diagram_capture/lib/diagram_capture.dart
```diff
@@ -206,7 +206,12 @@ class _DiagramFlutterBinding extends BindingBase
 
 /// A generator used by [DiagramController] for generating filenames for frames
 /// in animated diagram sequences.
-typedef File AnimationFilenameGenerator();
+typedef AnimationFilenameGenerator = File Function();
+
+/// A keyframe function designed to be executed before a particular frame is
+/// captured. Passes the current value of "now" (the time elapsed since the
+/// beginning of the animation).
+typedef DiagramKeyframe = void Function(Duration duration);
```
Hixie: i think we're still using the old style, i don't think we've enabled the new lint yet

diagrams/lib/transitions.dart
```diff
@@ -0,0 +1,599 @@
...
+  AnimationController _controller;
+
+  @override
+  void initState() {
+    super.initState();
+    _controller = new AnimationController(
+      duration: _kAnimationDuration,
+      vsync: this,
+    )..addListener(() {
+        setState(() {});
```
Hixie: always put a comment inside the `{}` of an empty setstate saying what changed (see style guide, i think it mentions that)

diagrams/lib/transitions.dart
```diff
@@ -0,0 +1,599 @@
...
+  }
+
+  @override
+  void dispose() {
+    _controller?.dispose();
+    super.dispose();
+  }
+
+  Widget wrap(BuildContext context, Widget child) {
+    return GestureDetector(
```
Hixie: i think we're still giving the "new", though the lint to stop that is coming soon i hope

diagrams/lib/transitions.dart
```diff
@@ -0,0 +1,599 @@
...
+    end: const BoxDecoration(
+      borderRadius: _endRadius,
+      color: const Color(0xffffffff),
+    ),
+  );
+
+  @override
+  DecoratedBoxTransitionDiagramState createState() {
+    return new DecoratedBoxTransitionDiagramState();
+  }
```
Hixie: since every state is just a build method, consider just putting the build method in these widgets and having the superclass state call it.

Hixie:
LGTM

PR #39, opened 2018-07-27
Rearrange the repo, making it conform to a more "standard" layout.

gspencergoog:
I've done the following here:

- Moved the `generate.dart` into the `bin` directory.
- Moved the testing script to the `bin` directory (and renamed it `testing.sh`)
- Added a script that will do the equivalent of update-packages
- Made the `animation_metadata` into a separate package, so that the generate.dart script and the diagram_capture library can both depend upon it (it needed to not depend upon Flutter).
- Moved the "diagrams" into `packages` and made it into a real package with exports.
  - This is in preparation for making a small utility that can run the diagrams interactively for development testing.
  - required renaming some globals to be privates so they wouldn't collide.
- Updated the testing script to point to the new locations.
- Fixed `generate_test.dart` so it passes again.
- Fixed a deprecation problem with `fake_process_manager.dart`, that was still using `typed`, and no longer needs to.
- Update README.md files.
- Switch to Cirrus CI for testing.


packages/diagrams/lib/diagrams.dart
```diff
@@ -0,0 +1,22 @@
...
+library diagrams;
+
+export 'src/app_bar.dart';
+export 'src/blend_mode.dart';
+export 'src/box_fit.dart';
+export 'src/card.dart';
+export 'src/colors.dart';
+export 'src/curve.dart';
+export 'src/diagram_step.dart';
+//export 'src/implicit_animations.dart';
```
Hixie: commented out code

gspencergoog: Fixed

PR #40, opened 2018-08-01
Implicit animations

gspencergoog:
This factors out a set of "animation diagram" base classes to make it easier to add the implicit animation diagrams.  I was going to have them all be based off the same class, but I realized that implicit animations (by design) have no way to expose the controller for driving the curve sparklines, so they didn't have enough in common to make that happen.

But I did make animated diagrams for the following classes:

- AnimatedAlign
- AnimatedContainer
- AnimatedDefaultTextStyle
- AnimatedOpacity
- AnimatedPadding
- AnimatedPhysicalModel
- AnimatedPositioned
- AnimatedPositionedDirectional
- AnimatedTheme


packages/diagrams/lib/src/animation_diagram.dart
```diff
@@ -0,0 +1,315 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:ui' as ui;
+
+import 'package:flutter/material.dart';
+
+import 'diagram_step.dart';
+
+const Duration _kAnimationDuration = const Duration(seconds: 2);
```
tvolkert: implicit `const`?

packages/diagrams/lib/src/animation_diagram.dart
```diff
@@ -0,0 +1,315 @@
...
+  String get name => _getName(runtimeType) + (decorate ? '' : '_plain');
+
+  String get caption => _getCaption(runtimeType);
+
+  @override
+  AnimationDiagramState<T> createState() => new AnimationDiagramState<T>();
+}
+
+class AnimationDiagramState<T> extends State<AnimationDiagram<T>> //
+    with
```
tvolkert: Weird formatting here

packages/diagrams/lib/src/animation_diagram.dart
```diff
@@ -0,0 +1,315 @@
...
+      if (match.start != 0) {
+        return '_${match.group(1).toLowerCase()}';
+      } else {
+        return match.group(1).toLowerCase();
+      }
+    },
+  );
+}
+
+abstract class AnimationDiagram<T> extends StatefulWidget implements DiagramMetadata {
```
tvolkert: Worth documenting the public members and functions for this?

PR #43, opened 2018-08-23
add a heroes terminology diagram

xster:
For https://github.com/flutter/flutter/pull/20322


gspencergoog:
I'd prefer if this were code, but I understand that it's non-trivial to make.  Have you run optpng on the image with the parameters that are in the README.md?

The pubspecs in this repo appear to be out of date. I sent you a PR to review that will fix that so your PR will go green: https://github.com/flutter/assets-for-api-docs/pull/44

xster:
Ya unlike app bars etc, in this case, I unfortunately wouldn't end up using the Hero or anything related to it to create the diagram. The code would just be a WYSIWYM drawer.

Hixie:
This should be code, even if it's just canvas calls or SVG or some such. That way we can easily update it, translate it, etc.

Hixie:
(or, probably most relevant, so that we can eventually make higher-res versions of it easily)

xster:
Ok, I swapped out the image for code. 

Tried looking for a WYSIWYM diagram editor but didn't find any good ones.

PR #44, opened 2018-08-23
Update pubspecs to work with dart 2.1.0

gspencergoog:
This updates the pubspecs to work with Dart 2.1.0, and moves one of the examples up a directory, since pub and flutter forbid examples with more than one example in them. Luckily I only had one...


xster:
RSLGTM

PR #48, opened 2018-12-15
WIP TextFormField Diagrams

justinmc:
This PR will create two diagrams: focused and unfocused TextFormField.


justinmc:
It was somewhere in the back of my mind, but this is getting pretty old.  I'll wait for the new diagrams and check in on this later.  Thanks for the diagram work!

PR #59, opened 2019-03-09
Container Diagrams

Piinks:
This PR adds generation code for container diagrams for the API docs. ⬛️


packages/diagrams/lib/src/container.dart
```diff
@@ -0,0 +1,81 @@
+// Copyright 2017 The Chromium Authors. All rights reserved.
```
gspencergoog: 2019

PR #60, opened 2019-03-14
ListView Diagrams

Piinks:
This PR adds generation code for container diagrams for the API docs. 📋


packages/diagrams/lib/src/list_view.dart
```diff
@@ -0,0 +1,108 @@
...
+  Widget build(BuildContext context) {
+    Widget returnWidget;
+    switch (name) {
+      case 'list_view':
+        returnWidget = ListView(
+          padding: const EdgeInsets.all(8.0),
+          children: <Widget>[
+            Container(
+              height: 50,
+              color: Colors.amber[600],
```
liyuqian: `colorCodes[0]`?

Also, I think it would be shorter to have:
```
children: <int>[0, 1, 2].map((i) => Container(
  height: 50, 
  color: Colors.amber[colorCodes[i]], 
  child: const Center(child: Text('Entry ${entries[index]}')).toList()
```

It would be even shorter if we define
```
  static final List<Widget> entries = <String>['A', 'B', 'C'].map((t) => const Center(child: Text('Entry $t'))).toList();
  static final List<Color> colors = <int>[600, 500, 100].map((i) => Colors.amber[i]).toList();
```
which can be used in the other 2 places.

Piinks: Thanks for the feedback! These are meant to match the sample code in the API docs, that's why they are a little less conventional. You can see the samples [here](https://master-api.flutter.dev/flutter/widgets/ListView-class.html).

liyuqian:
I see. Thanks for improving the API docs!

PR #61, opened 2019-03-15
Align Diagrams

Piinks:
## Description

This PR contains generator code and diagrams for the Align class. 
Diagrams based on snippets from [flutter/flutter/pull/29442](https://github.com/flutter/flutter/pull/29442).

## Related Issues

Part (2/3) of [flutter/flutter/issues/13637](https://github.com/flutter/flutter/issues/13637) : Add images and examples to Align
Will need follow-up of:
* Embedding of diagrams into API docs from [flutter/flutter](https://github.com/flutter/flutter)


packages/diagrams/lib/src/align.dart
```diff
@@ -0,0 +1,114 @@
...
+    const Widget logo = FlutterLogo(size: 60);
+    const Icon origin = Icon(
+      Icons.fiber_manual_record,
+      size: 20,
+    );
+    Widget heading;
+    Widget containerChild;
+    switch (name) {
+      case 'align_constant':
+        heading = const Text('Alignment.topRight');
```
gspencergoog: You can just use this, and then it'll automatically change if the name of the classes change:
```suggestion
        heading = const Text(Alignment.topRight.toString());
```

packages/diagrams/lib/src/align.dart
```diff
@@ -0,0 +1,114 @@
...
+        containerChild = const Align(
+          alignment: Alignment.topRight,
+          child: logo,
+        );
+        break;
+      case 'align_alignment':
+        heading = const Text('Alignment Origin');
```
gspencergoog: Shouldn't this be "Alignment.center"? What are you trying to illustrate here, the center alignment constant, or the location of the origin?

packages/diagrams/lib/src/align.dart
```diff
@@ -0,0 +1,114 @@
...
+
+  @override
+  final String name;
+
+  @override
+  Widget build(BuildContext context) {
+//    List<Widget> diagramChildren;
+    const Widget logo = FlutterLogo(size: 60);
+    const Icon origin = Icon(
+      Icons.fiber_manual_record,
```
gspencergoog: I think I'd use something like "gps_fixed" here instead of just a dot.

Also, for the Fractional Offset Origin diagram, the dot doesn't appear on the upper corner of the container, it is inside of the padding, so it's hard to know that the origin is the top left of the container.

packages/diagrams/lib/src/align.dart
```diff
@@ -0,0 +1,114 @@
...
+            Align(
+              alignment: Alignment.center,
+              child: origin,
+            ),
+          ],
+        );
+
+        break;
+      case 'align_fractional_offset':
+        heading = const Text('Fractional Offset Origin');
```
gspencergoog: I'm also a little confused about the context of this diagram: is it supposed to show where the origin is for the fractional offset?

The location of the logo looks kind of arbitrary too.

Piinks:
Ah thanks!
These diagrams are intended to go alongside some new sample snippets that just landed today, [here](https://github.com/flutter/flutter/pull/29442), and in the docs [here](https://master-api.flutter.dev/flutter/widgets/Align-class.html).
The dots (I'll change them to 'gps_fixed', that's much better!) are meant to show the differing origins for FractionalOffset and Alignment, while the logo shows how the same parameters reflect different positions in accordance with each coordinate system.
It is complicated - I've been trying to illustrate it in a way that is simple. Does the explanation in the docs work alongside these images, or should I go back to the drawing board?

packages/diagrams/lib/src/align.dart
Piinks: The location of the origin is the intention here. FractionlOffset and Alignemnt have different origins.

PR #65, opened 2019-04-02
Run pngquant over all pngs

kevmoo:
Dropped ~4MB

`find . -name '*.png' -print0 | xargs -0 -P8 -L1 pngquant --ext .png --force --speed 1 --skip-if-larger`


gspencergoog:
Huh.  So should we switch to using `pngquant` instead of `optipng`?  Otherwise the sizes will all go back up once we run the generation script again.

gspencergoog:
Or, I suppose we could run them both, and let them fight it out to see which one makes the smaller image.

kevmoo:
> Huh. So should we switch to using `pngquant` instead of `optipng`? Otherwise the sizes will all go back up once we run the generation script again.

Or switch to pngquant – or turn up the volume.

consider this more of an FYI – smaller is better for the host, user, etc...

kevmoo:
See also https://github.com/flutter/flutter/pull/30397

gspencergoog:
OK, thanks. The "volume" on the `optipng` that we're running is already turned all the way up:
```bash
optipng -zc1-9 -zm1-9 -zs0-3 -f0-5
```

Yeah, smaller is better (unless quality drops), which is why we were running `optipng` on everything already.

The quantization might be problematic for some images (smooth gradients, lots of colors), so we might need a blacklist, but overall it seems like a win.

gspencergoog:
When I run it, it only drops about 1MB for me on the assets directory: the other PNGs are not served, they're only used to build the tool, so their size doesn't really matter.

gspencergoog:
I'll take a look at adding it to the process.

kevmoo:
> When I run it, it only drops about 1MB for me on the assets directory: the other PNGs are not served, they're only used to build the tool, so their size doesn't really matter.

Ah. Yeah, my number was for the whole repo.

Feel free to close this out, take inspiration, whatever...

PR #71, opened 2019-05-01
Flow Layout Widget

Piinks:
This PR adds an animation that shows a menu created using the `Flow` widget. It matches the sample code that will be in the API docs for this widget.
This is part of fixing issue [#10146](https://github.com/flutter/flutter/issues/10146) in flutter/flutter.


packages/diagrams/lib/src/flow.dart
```diff
@@ -0,0 +1,172 @@
+import 'dart:async';
```
Hixie: all dart files need a license block

packages/diagram_capture/lib/diagram_capture.dart
```diff
@@ -441,6 +442,8 @@ class DiagramController {
         }
       }
 
+      if(gestureCallback != null)
```
Hixie: trivial nit: put a space between keywords/identifiers and `(` unless it's a method call

gspencergoog:
Also, you'll probably need to fix these to get the analysis to pass, even though you didn't change that file.
```
  info • Prefer const with constant constructors • diagrams/lib/src/transitions.dart:232:12 • prefer_const_constructors
   info • Prefer const with constant constructors • diagrams/lib/src/transitions.dart:233:10 • prefer_const_constructors
```

PR #72, opened 2019-05-06
Tab/TabBar/TabBarView Widget Animation

shihaohong:
Adds an animation of the code sample for `TabBar`, `TabBarView` and `Tabs`. This helps resolve https://github.com/flutter/flutter/issues/31392


packages/diagrams/lib/src/tabs.dart
```diff
@@ -0,0 +1,124 @@
...
+  void tapTabs(DiagramController controller, Duration now) async {
+    RenderBox target;
+    switch(now.inMilliseconds) {
+      case 0:
+        target = _tabKeys[1].currentContext.findRenderObject();
+        break;
+      case 1300:
+        target = _tabKeys[0].currentContext.findRenderObject();
+        break;
+      default:
```
Piinks: Might be worth adding a default, just in case.

packages/diagrams/lib/src/tabs.dart
```diff
@@ -0,0 +1,124 @@
...
+import 'dart:async';
+import 'dart:io';
+
+import 'package:diagram_capture/diagram_capture.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+import 'diagram_step.dart';
+
+const Duration _kTabScrollDuration = Duration(milliseconds: 300);
```
Piinks: Uber nit here & below in naming - there is no scrolling in this animation

packages/diagrams/lib/src/tabs.dart
```diff
@@ -0,0 +1,124 @@
...
+  @override
+  final String name;
+
+  @override
+  State<TabsDiagram> createState() => TabsDiagramState();
+}
+
+class TabsDiagramState extends State<TabsDiagram> with SingleTickerProviderStateMixin {
+  final List<Tab> myTabs = <Tab>[
+    Tab(key: _tabKeys[0], text: 'Left'),
```
Piinks: This looks great, the animation is very subtle though. It may be worth adding something on the left and right pages that are different. This would make the change more apparent than the little navigation cues in the tab bar.

packages/diagrams/lib/src/tabs.dart
```diff
@@ -0,0 +1,124 @@
...
+    controller.builder = (BuildContext context) => typedDiagram;
+    return await controller.drawAnimatedDiagramToFiles(
+      end: _kTotalAnimationTime,
+      frameRate: _kAnimationFrameRate,
+      name: diagram.name,
+      category: category,
+      gestureCallback: tapTabs,
+    );
+  }
+}
```
gspencergoog: nit: Add a newline..

packages/diagrams/lib/src/tabs.dart
shihaohong: The TabView animation is actually a scroll animation under the hood (I got too used to it from working on it recently). However, I do agree that it can be vague without knowing this. Would `tabChangeDuration` or `tabSwitchDuration` be better alternatives?

packages/diagrams/lib/src/tabs.dart
shihaohong: I agree, I'll look into a way to highlight the tab change a little better.

packages/diagrams/lib/src/tabs.dart
shihaohong: What would you suggest adding? I want to avoid random tapping of tabs if it doesn't hit either case 0 or 1300

Piinks: Oh you are right, good point.

packages/diagrams/lib/src/tabs.dart
Piinks: Oh fun fact! It's not a huge deal. You can leave it if you prefer.

PR #76, opened 2019-05-13
Add code to generate loading progress image animation.

tvolkert:
flutter/flutter#32374


tvolkert:
I'm going to remove the stuff that's specific to the loading progress APIs since those APIs got reworked a bit.  But the other stuff in here is still very useful and can be merged today, so proceeding with that.  I'll send a separate PR for the loading progress stuff once the flutter/flutter PR lands.

PR #78, opened 2019-05-16
Restore the checkerboard pattern to blend mode diagrams.

Hixie:
cc @gspencergoog


README.md
```diff
@@ -20,29 +20,24 @@ material library would go in the `assets/material/` directory and be at
...
+```sh
+PATH=~/dev/flutter/bin/cache/dart-sdk/bin:~/Android/Sdk/platform-tools:$PATH bin/generate.sh
+```
 
-See the [documentation for the generate.dart script in the `diagrams`
-directory](packages/diagrams/README.md), in conjunction with the
-[`generate.dart`](./bin/generate.dart) script. It will regenerate almost all of
-existing assets using the Flutter version you have installed. Feel free
-to add more modules in the `diagrams` package to generate new assets.
+You may wish to add the `--category` argument to focus just on the
```
gspencergoog: You might mention `--name`, which is almost more useful than `--category`.

packages/diagrams/lib/src/blend_mode.dart
```diff
@@ -16,9 +16,9 @@ import 'diagram_step.dart';
 Completer<Null> touch;
 final GlobalKey key = new GlobalKey();
 
-const String destinationImageName = 'assets/blend_mode_destination.jpeg';
-const String sourceImageName = 'assets/blend_mode_source.png';
-const String gridImageName = 'assets/blend_mode_grid.png';
+const ImageProvider destinationImageProvider = const ExactAssetImage('assets/blend_mode_destination.jpeg', package: 'diagrams');
```
gspencergoog: I don't think you need the second const here.

README.md
Hixie: I couldn't figure out how to use `--name`.

packages/diagrams/lib/src/blend_mode.dart
Hixie: oops.

Hixie:
> None of the blend mode images had any checkerboard for me. They were all over a transparent background, but with no checkerboard, and the parrot and background image were missing in all of them.

With this patch? Or before?
They were broken in https://github.com/flutter/assets-for-api-docs/commit/2abe9a42c3901fd0831f8e965a303967d780a9a6#diff-debf4c1f27e96284386dea7ef00a0777

gspencergoog:
In this patch.

Hixie:
Ok that's weird. I ran it again and it seemed to fix it... I don't understand. It may have been because of my doing weird things with git earlier to not include the mp4 file updates, which I didn't bother with this time. (The files look identical to me but apparently aren't bit-for-bit identical.)

Hixie:
cc @Piinks for review since @gspencergoog is away.

README.md
Piinks: `--category` can narrow down the generation script to a library of diagrams, while `--name` can drill down on a specific diagram. Every `DiagramMetadata` implementation has this name property which also serves as the file name. 
[Example Here](https://github.com/flutter/assets-for-api-docs/blob/951c8c526cc57c9634088971ae56c5197f65dc99/packages/diagrams/lib/src/switch_list_tile.dart#L188)

PR #85, opened 2019-06-06
Adding code and diagram for MediaQuery EdgeInsets

Piinks:
Code and generated diagram for https://github.com/flutter/flutter/pull/33999


packages/diagrams/lib/src/media_query.dart
```diff
@@ -0,0 +1,142 @@
...
+                    padding: const EdgeInsets.fromLTRB(240.0, 100.0, 0.0, .0),
+                    child: new Container(
+                      width: 6.0,
+                      height: 150.0,
+                      color: Colors.black45,
+                    ),
+                  ),
+                  const Padding(
+                    padding: EdgeInsets.fromLTRB(240.0, 260.0, 0.0, 0.0),
+                    child: Text(
```
dnfield: For these, I think there's a Label widget or something we can use that might look clearer.

Piinks: Using the label will draw a line directly to the text, which would go outside the bounds on the actual area I want to specify. Do you think that would be ok, or more confusing?

PR #87, opened 2019-06-19
Add text height diagram

GaryQian:
New simplified version that focuses on showing the relevant differences between the default font height and the em-square/font size height.

For https://github.com/flutter/flutter/pull/33281

This should be able to be landed cleanly before the height changes are in the engine.


packages/diagrams/lib/src/text.dart
```diff
@@ -0,0 +1,204 @@
...
+        text:' Aalg我हिन्दी ',
+        style: TextStyle(
+          fontSize: 100,
+          color: Colors.black,
+        ),
+      ),
+      textAlign: TextAlign.left,
+      textDirection: TextDirection.ltr,
+    );
+    switch (name) {
```
goderbauer: Why this switch?

packages/diagrams/lib/src/text.dart
```diff
@@ -0,0 +1,204 @@
...
+  const CurvePainter(this.textPainter);
+
+  final TextPainter textPainter;
+
+  @override
+  void paint(Canvas canvas, Size size) {
+    textPainter.layout();
+
+    final List<TextBox> boxes = textPainter.getBoxesForSelection(
+      const TextSelection(baseOffset: 0, extentOffset: 99)
```
goderbauer: What's the 99 magic number?

packages/diagrams/lib/src/text.dart
```diff
@@ -0,0 +1,204 @@
...
+      paint,
+    );
+
+    // emBottom
+    canvas.drawLine(
+      baseOffset + Offset(0, emBottom),
+      baseOffset + Offset(width, emBottom),
+      paint,
+    );
+
```
goderbauer: remove some of this extra white space?

packages/diagrams/lib/src/text.dart
```diff
@@ -0,0 +1,204 @@
...
+      textAlign: TextAlign.left,
+      textDirection: TextDirection.ltr,
+    );
+    label.layout();
+    label.paint(canvas, baseOffset + Offset(0, baseline + 3));
+  }
+
+  @override
+  bool shouldRepaint(CustomPainter oldDelegate) {
+    return false;
```
goderbauer: Don't you want to repaint if the textPainters are different?

packages/diagrams/lib/src/text.dart
```diff
@@ -0,0 +1,204 @@
+// Copyright 2019 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
```
goderbauer: group and separate imports per our style guide.

packages/diagrams/lib/src/text.dart
GaryQian: This allows for adding additional diagrams using the same the same class in the future based off of a different name. It is not being used currently, so I can remove this for now if you want.

packages/diagrams/lib/src/text.dart
GaryQian: It just needs to be a large number to ensure it extends out beyond the end of short string.

packages/diagrams/lib/src/text.dart
goderbauer: I'd just remove it and add that when we actually need it.

PR #89, opened 2019-06-27
Diagrams for "Top disliked pages"

brianegan:
Heya! I got a list of "Top Disliked API Docs pages" from Emily, and thought I'd start the improvement by adding Diagrams for these pages.

Most of these are pretty straight-forward, but I had one question about how to handle the `Image` Widget API docs. 

Images load asynchronously, so it appears we cannot take a normal screenshot and continue on (since the image might not be loaded). To capture Diagrams of the `Image` widget and it's different constructors, we need to  record a video instead. 

However, that didn't seem like it would be the best experience folks reading the documentation, since the videos they'd see start with a blank white canvas and then the image pops in. Therefore, I've simply included the `owl.jpg` found on the Flutter site in this repo as well so we can use that image as the "Demo" in the API docs for each of the constructors.

For normal constructors, that's no biggie. However, if we wanted to demonstrate some of the more advanced features of  the `Image` Widget, such as `colorBlendMode`, this option might not be the best and the videos might be better. Looking forward to any feedback in this regard!


Piinks:
> Images load asynchronously, so it appears we cannot take a normal screenshot and continue on (since the image might not be loaded). To capture Diagrams of the `Image` widget and it's different constructors, we need to record a video instead.
> 
> For normal constructors, that's no biggie. However, if we wanted to demonstrate some of the more advanced features of the `Image` Widget, such as `colorBlendMode`, this option might not be the best and the videos might be better. Looking forward to any feedback in this regard!

Good point!
I would check the colorBlend diagrams that are already in this repo. They may offer some guidance. You can see them in the API Docs here: https://api.flutter.dev/flutter/dart-ui/BlendMode-class.html

brianegan:
Thanks for the feedback! 

Yah, I'd looked at the colorBlend images as well -- perhaps it's overkill to demonstrate the `colorBlendMode` in the `Image` API docs themselves. If folks wonder how those work, linking directly to that page might be the best option since it is a nice visual breakdown of the various options.

I've patched up the analysis issues and therefore think this is ready to merge pending any other feedback!

packages/diagrams/lib/src/scaffold.dart
```diff
@@ -0,0 +1,66 @@
...
+  @override
+  State<StatefulWidget> createState() => ScaffoldDiagramState();
+}
+
+class ScaffoldDiagramState extends State<ScaffoldDiagram> {
+  int _count = 0;
+
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
```
Piinks: I might put this in a constrained box or something similar to enforce a typical size and layout.

This diagram could also be incredibly informative by using some of the other features of the Scaffold class and labelling them. You could even create more than one diagram to illustrate the different features as well!

packages/diagrams/lib/src/column.dart
```diff
@@ -0,0 +1,102 @@
...
+
+const String _column = 'column';
+const String _columnWithProperties = 'column_properties';
+
+class ColumnDiagram extends StatelessWidget implements DiagramMetadata {
+  const ColumnDiagram(this.name);
+
+  @override
+  final String name;
+  static const List<String> entries = <String>['A', 'B', 'C'];
```
Piinks: Where are these `entries` or `colorCodes` being used?

packages/diagrams/lib/src/column.dart
```diff
@@ -0,0 +1,102 @@
...
+        break;
+    }
+    return new ConstrainedBox(
+      key: new UniqueKey(),
+      constraints: new BoxConstraints.tight(const Size(400.0, 250.0)),
+      child: new Container(
+          alignment: FractionalOffset.center,
+          padding: const EdgeInsets.all(5.0),
+          color: Colors.white,
+          child: returnWidget),
```
Piinks:
```suggestion
          child: returnWidget,
        ),
```

packages/diagrams/lib/src/column.dart
```diff
@@ -0,0 +1,102 @@
...
+        );
+        break;
+      case _columnWithProperties:
+        returnWidget = Column(
+          crossAxisAlignment: CrossAxisAlignment.start,
+          mainAxisSize: MainAxisSize.min,
+          children: <Widget>[
+            const Text('We move under cover and we move as one'),
+            const Text(
+              'Through the night, we have one shot to live another day',
```
Piinks: Nits: These text excerpts can all be on the same line.

packages/diagrams/lib/src/column.dart
```diff
@@ -0,0 +1,102 @@
...
+            const Text(
+              'We will fight up close, seize the moment and stay in it',
+            ),
+            const Text(
+              'It’s either that or meet the business end of a bayonet',
+            ),
+            const Text('The code word is ‘Rochambeau,’ dig me?'),
+            Text(
+              'Rochambeau!',
+              style:
```
Piinks: Nit: This can be on the same line.

packages/diagrams/lib/src/row.dart
```diff
@@ -0,0 +1,111 @@
...
+const String _row = 'row';
+const String _rowError = 'row_error';
+const String _rowFixed = 'row_fixed';
+
+class RowDiagram extends StatelessWidget implements DiagramMetadata {
+  const RowDiagram(this.name);
+
+  @override
+  final String name;
+  static const List<String> entries = <String>['A', 'B', 'C'];
```
Piinks: Same comment as before for `entries` and `colorCodes`

packages/diagrams/lib/src/stack.dart
```diff
@@ -0,0 +1,112 @@
...
+
+const String _stack = 'stack';
+const String _stackWithGradient = 'stack_with_gradient';
+
+class StackDiagram extends StatelessWidget implements DiagramMetadata {
+  const StackDiagram(this.name);
+
+  @override
+  final String name;
+  static const List<String> entries = <String>['A', 'B', 'C'];
```
Piinks: Same comment as before for `entries` and `colorCodes`

packages/diagrams/lib/src/text.dart
```diff
@@ -54,22 +122,23 @@ class TextDiagramPainter extends CustomPainter {
...
-    );
+        const TextSelection(baseOffset: 0, extentOffset: largeIndex));
 
     final Paint paint = Paint();
     paint.color = Colors.black;
     paint.strokeWidth = 3.5;
     const double top = 0;
     final double bottom = textPainter.height;
-    final double baseline = textPainter.computeDistanceToActualBaseline(TextBaseline.alphabetic);
+    final double baseline =
```
Piinks: Nit: This can be one line

packages/diagrams/lib/src/scaffold.dart
brianegan: Good call, fixed!

brianegan:
Thanks so much for the review @Piinks and apologies for the delay on this one -- busy week in Berlin helping the Flutter team at Droidcon :) Please let me know if you see any further issues or would like to see specific examples I might have missed for any of these!

A couple of questions:

 1. I've added a couple more screenshots to the Scaffold class, but it seemed like some of the possible examples fit be better in their respective API docs, such as more complex BottomAppBar or BottomSheet kind of examples.
  2. I've fixed all of the nits, but also thought it might be a good idea to push a PR that formats the codebase with `dartfmt` by default. We can always ignore `dartfmt` for the code samples in the API docs to make them as readable as possible, but for this codebase it might be easier if we all use a common format. 

Thoughts?

Piinks:
Looking good so far!
Looks like there is a small nit from the analyzer:

- info • Private field could be final • diagram_capture/lib/diagram_capture.dart:137:26 • prefer_final_fields

Once that is fixed this should be good to go! 👍 

As for your questions:
> 1. I've added a couple more screenshots to the Scaffold class, but it seemed like some of the possible examples fit be better in their respective API docs, such as more complex BottomAppBar or BottomSheet or Drawer kind of examples. Do you see any other options that might make sense for the Scaffold docs?

That's a great idea! Perhaps on the future PR in flutter/flutter where we'll be embedding these diagrams, we can make sure that there is clear guidance to the relevant resources for further discovery. :) 

> 2. I've fixed all of the nits, but also thought it might be a good idea to push a PR that formats the codebase with `dartfmt` by default. We can always ignore `dartfmt` for the code samples in the API docs to make them as readable as possible, but for this codebase it might be easier if we all use a common format. Thoughts?

For the Flutter codebase we don't use `dartfmt`. We have a common format that is laid out by the [Flutter Style Guide](https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo).

brianegan:
Sounds great, thanks again for the review! I think all issues have been resolved and merged with latest master changes to the text diagrams

brianegan:
Woohoo! Thanks @Piinks :) Sorry about the delay -- just back from a short vacation.

PR #90, opened 2019-07-08
New text height diagram, update old ones.

GaryQian:
This adds additional diagrams to fully illustrate the behavior of TextStyle.height after the beraking change in https://github.com/flutter/engine/pull/9041

The goal is to show that `height:null != height:1.0`

These diagrams should be regenerated after the change lands, as they are currently reflecting the behavior without the change.


GaryQian:
Analyzer is fixed :)

PR #93, opened 2019-07-18
Add text border example

GaryQian:
This adds an example of using foreground to create a fill and stroke text.


dnfield:
Nice!  Do you think you could add an example of doing shaded text as well (e.g. text with a Linear or Radial gradient or something)?

GaryQian:
Ok, I'll do it in a followup PR!

PR #97, opened 2019-07-23
Next 9 "Top" Diagrams

brianegan:
Hey hey :)

I've added diagrams for the following Widgets / Classes:

  - [Icons](https://master-api.flutter.dev/flutter/material/Icons-class.html)
  - [MaterialApp](https://master-api.flutter.dev/flutter/material/MaterialApp-class.html)
  - [TextStyle](https://master-api.flutter.dev/flutter/painting/TextStyle-class.html)
  - [FlatButton](https://master-api.flutter.dev/flutter/material/FlatButton-class.html)
  - [Expanded](https://master-api.flutter.dev/flutter/widgets/Expanded-class.html)
  - [GridView](https://master-api.flutter.dev/flutter/widgets/GridView-class.html)
  - [FloatingActionButton](https://master-api.flutter.dev/flutter/material/FloatingActionButton-class.html)
  - [DropdownButton](https://master-api.flutter.dev/flutter/material/DropdownButton-class.html)
  - [CustomScrollView](https://master-api.flutter.dev/flutter/widgets/CustomScrollView-class.html)

I tried to record videos for the DropdownButton and AlertDialog (not included) Widgets, but the diagram capture process only takes screenshots of a specific Widget, and therefore it does not capture the new Widgets / Routes placed above these Widgets.

If there's way / advice on how to capture that kind of interaction, please let me know!


brianegan:
Thanks @Piinks! I've updated the description to add links for each Diagram to the corresponding API doc.

I've also taken a look at the `Flow` diagram and unfortunately that technique also wouldn't quite work. It appears as though the diagram generator only takes a picture (or video frames) of the Widget returned by the `DiagramController.builder` method. Since some of these Widgets launch a new Widget using the navigator, the tool cannot take pictures of those at this time.

I'll take a look at the generator a bit more to see if I can tweak it to work with Navigator-based classes.

brianegan:
Thanks so much for the thorough review, @Piinks! I've updated all of the examples based on your feedback -- it all made a lot of sense.

Please let me know if you like the updates or spot anything else that could use some more attention :)

packages/diagrams/lib/src/flat_button.dart
```diff
@@ -0,0 +1,85 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    Widget returnWidget;
+
+    switch (name) {
+      case _basic:
+        returnWidget = FlatButton(
+          onPressed: () {
+            /*...*/
```
Piinks: Comments from the code samples don't need to be included here.

packages/diagrams/lib/src/flat_button.dart
```diff
@@ -0,0 +1,85 @@
...
+      case _properties:
+        returnWidget = FlatButton(
+          color: Colors.blue,
+          textColor: Colors.white,
+          disabledColor: Colors.grey,
+          disabledTextColor: Colors.black,
+          padding: const EdgeInsets.all(8.0),
+          splashColor: Colors.blueAccent,
+          onPressed: () {
+            /*...*/
```
Piinks: Here as well.

packages/diagrams/lib/src/expanded.dart
```diff
@@ -0,0 +1,119 @@
...
+
+  @override
+  final String name;
+
+  @override
+  Widget build(BuildContext context) {
+    Widget returnWidget;
+
+    switch (name) {
+      case _column:
```
Piinks: The column appears to be filling the entire viewport, while the row has a fixed height of 100. Let's make the column a fixed width of 100 to match.
Also, let's make the expanded/not expanded colors the same between examples to better illustrate which is which.

packages/diagrams/lib/src/expanded.dart
```diff
@@ -0,0 +1,119 @@
...
+          body: Center(
+            child: Column(
+              children: <Widget>[
+                Container(
+                  color: Colors.green,
+                  height: 100,
+                ),
+                Expanded(
+                  child: Container(
+                    color: Colors.yellow,
```
Piinks:
```suggestion
                    color: Colors.amber,
```

packages/diagrams/lib/src/expanded.dart
```diff
@@ -0,0 +1,119 @@
...
+                    child: const Center(child: Text('Expanded')),
+                  ),
+                ),
+                Container(
+                  color: Colors.green,
```
Piinks:
```suggestion
                  color: Colors.blue,
```

packages/diagrams/lib/src/expanded.dart
```diff
@@ -0,0 +1,119 @@
...
+      case _column:
+        returnWidget = Scaffold(
+          appBar: AppBar(
+            title: const Text('Expanded Column Sample'),
+          ),
+          body: Center(
+            child: Column(
+              children: <Widget>[
+                Container(
+                  color: Colors.green,
```
Piinks:
```suggestion
                  color: Colors.blue,
```

packages/diagrams/lib/src/expanded.dart
```diff
@@ -0,0 +1,119 @@
...
+          ),
+          body: Center(
+            child: Row(
+              children: <Widget>[
+                Expanded(
+                  flex: 2,
+                  child: Container(
+                    color: Colors.amber,
+                    height: 100,
+                    child: const Text('Expanded(\n  flex: 2\n)'),
```
Piinks: Let's see if we can center, or create a label widget.

packages/diagrams/lib/src/expanded.dart
```diff
@@ -0,0 +1,119 @@
...
+                  color: Colors.blue,
+                  height: 100,
+                  width: 50,
+                ),
+                Expanded(
+                  flex: 1,
+                  child: Container(
+                    color: Colors.amber,
+                    height: 100,
+                    child: const Text('Expanded(\n  flex: 1\n)'),
```
Piinks: Same here.

packages/diagrams/lib/src/floating_action_button.dart
```diff
@@ -0,0 +1,93 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    Widget returnWidget;
+
+    switch (name) {
+      case _basic:
+        returnWidget = Scaffold(
+          appBar: AppBar(
+            title: const Text('Floating Action Button Sample'),
```
Piinks:
```suggestion
            title: const Text('Floating Action Button'),
```
Looks like the 'Sample` was being truncated in the diagram. We can just remove it so there isn't an ellipsis in the heading.

packages/diagrams/lib/src/floating_action_button.dart
```diff
@@ -0,0 +1,93 @@
...
+          ),
+          body: const Center(
+              child: Text('Press the button below!')
+          ),
+          floatingActionButton: FloatingActionButton(
+            onPressed: () {
+              // Add your onPressed code here!
```
Piinks: Also can be removed.

packages/diagrams/lib/src/floating_action_button.dart
```diff
@@ -0,0 +1,93 @@
...
+        returnWidget = Scaffold(
+          appBar: AppBar(
+            title: const Text('Floating Action Button Label'),
+          ),
+          body: const Center(
+            child: Text('Press the button with a label below!'),
+          ),
+          floatingActionButton: FloatingActionButton.extended(
+            onPressed: () {
+              // Add your onPressed code here!
```
Piinks: Same here.

packages/diagrams/lib/src/icon.dart
```diff
@@ -20,15 +22,32 @@ class IconDiagram extends StatelessWidget implements DiagramMetadata {
...
-          size: 30.0,
+        child: Center(
+          child: Row(
+            mainAxisAlignment: MainAxisAlignment.spaceAround,
+            children: const <Widget>[
+              Icon(
+                Icons.favorite,
+                color: Colors.pink,
+                size: 24.0,
+                semanticLabel: 'Text to announce in accessibility modes',
```
Piinks: This is great to include in the sample code for the docs, but we don't need it here for the static image generator.

packages/diagrams/lib/src/expanded.dart
brianegan: Oh my. For some reason I thought the word "Expanded" would run longer than 100 pixels wide, but that wasn't the case -- yet I made these changes anyway to make sure it fit and the column didn't blend with the app bar! Egg on my face, and 100 pixel width + colors returned to their natural state.

packages/diagrams/lib/src/expanded.dart
brianegan: Yah, this is slightly trickier -- I originally split the text on newlines because I felt it looks a bit awkward when the `Expanded(flex: 1)` text didn't fit on a single line. Here I've centered the text and removed the `\n` newlines and this is the result:

expanded_row

What do you think of this version? I didn't want to nest too many Widgets in here to keep the sample in the API docs simple and didn't want folks copy-pasting the sample to get a different output than what the image shows. Any feedback / ideas appreciated!

Piinks: I see what you mean. 

I would suggest trying to just include 'flex: n' and see how that looks, or make use of the label widget https://github.com/flutter/assets-for-api-docs/blob/3333f4a2832912c1e80220e5e053290f6c764374/packages/diagrams/lib/src/utils.dart#L37 that `utils.dart` provides for this type of thing.

We don't have to worry about the sample in the API docs getting too complicated, because we don't need to include things like labels in the actual snippet. We can use the diagrams to point out or illustrate the different elements and features, but not necessarily/explicitly include those visual aids in the actual snippet. Does that make sense?

packages/diagrams/lib/src/grid_view.dart
```diff
@@ -0,0 +1,96 @@
...
+
+    switch (name) {
+      case _gridView:
+        returnWidget = GridView.count(
+          primary: false,
+          padding: const EdgeInsets.all(20.0),
+          crossAxisSpacing: 10.0,
+          crossAxisCount: 2,
+          children: <Widget>[
+            Container(child: const Center(child: Text('He\'d have you all unravel at the')), color: Colors.teal[100]),
```
Piinks: Okay, apologies for the iteration. Maybe `Center` was not the best idea since some of these strings are longer than others, and the longer ones come right up to the top/left edges. How about instead:
```suggestion
            Container(child: const Padding(padding: EdgeInsets.all(8.0), child: Text('He\'d have you all unravel at the')), color: Colors.teal[100]),
```
How does that look?

packages/diagrams/lib/src/grid_view.dart
```diff
@@ -0,0 +1,96 @@
...
+        );
+        break;
+      case _customScrollGridView:
+        returnWidget = CustomScrollView(
+          primary: false,
+          slivers: <Widget>[
+            SliverPadding(
+              padding: const EdgeInsets.all(20.0),
+              sliver: SliverGrid.count(
+                crossAxisSpacing: 10.0,
```
Piinks: Let's add mainAxisSpacing as well.

packages/diagrams/lib/src/grid_view.dart
```diff
@@ -0,0 +1,96 @@
...
+  @override
+  Widget build(BuildContext context) {
+    Widget returnWidget;
+
+    switch (name) {
+      case _gridView:
+        returnWidget = GridView.count(
+          primary: false,
+          padding: const EdgeInsets.all(20.0),
+          crossAxisSpacing: 10.0,
```
Piinks: Let's add mainAxisSpacing as well.

packages/diagrams/lib/src/custom_scroll_view.dart
```diff
@@ -0,0 +1,166 @@
...
+        + _kScrollPauseDuration
+        + _kScrollDownDurationPartTwo
+        + _kScrollPauseDuration;
+const Duration _kScrollUpDuration = Duration(seconds: 1, milliseconds: 500);
+const Duration _kScrollDownDurationPartOne = Duration(milliseconds: 800);
+const Duration _kScrollDownDurationPartTwo = Duration(seconds: 1);
+const Duration _kScrollPauseDuration = Duration(milliseconds: 900);
+
+const double _kCurveAnimationFrameRate = 60.0;
+const String _customScrollView = 'custom_scroll_view';
```
Piinks: It looks like this diagram needs to be re-generated to update the latest changes.

packages/diagrams/lib/src/floating_action_button.dart
```diff
@@ -0,0 +1,89 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    Widget returnWidget;
+
+    switch (name) {
+      case _basic:
+        returnWidget = Scaffold(
+          appBar: AppBar(
+            title: const Text('Floating Action Button'),
```
Piinks: It looks like this sample needs to be run again to update the image above.

packages/diagrams/lib/src/text_style.dart
```diff
@@ -0,0 +1,133 @@
...
+              const TextSpan(
+                text: ' we got it made in the shade.',
+              ),
+            ],
+          ),
+        );
+        break;
+      case _customFonts:
+        returnWidget = const Text(
+          'Look, when Britain taxed our tea, we got frisky.',
```
Piinks: This image above also needs to be updated.

Piinks:
In case it is hidden above (Github does that sometimes 😝 ):
I see what you mean about the Expanded Row sample. 

I would suggest trying to just include 'flex: n' and see how that looks, or make use of the label widget https://github.com/flutter/assets-for-api-docs/blob/3333f4a2832912c1e80220e5e053290f6c764374/packages/diagrams/lib/src/utils.dart#L37 that `utils.dart` provides for this type of thing.

We don't have to worry about the sample in the API docs getting too complicated, because we don't need to include things like labels in the actual snippet. We can use the diagrams to point out or illustrate the different elements and features, but not necessarily/explicitly include those visual aids in the actual snippet. Does that make sense?

brianegan:
Thanks again for the thorough review, @Piinks! I think I've got everything updated based on your feedback. Please let me know if ya see any more room for improvement :)

brianegan:
No rush at all, but once these look good, I've opened a corresponding issue on the flutter repo to add these diagrams to the docs: https://github.com/flutter/flutter/pull/37624

brianegan:
Woohoo! Thanks again for all your help :)

PR #98, opened 2019-08-15
ToggleButtons assets

shihaohong:
Adds assets to display different ways that toggle buttons can be implemented.

Fixes https://github.com/flutter/flutter/issues/36862


shihaohong:
Did you mean to reduce the tap duration? That crossed my mind and I experimented with 500ms, 750ms and 1 second (I ended up going with 750ms). I was thinking that it would help emphasize the tap behavior when developers look at it in the docs, but I don't have a preference either way.

Piinks:
> Did you mean to reduce the tap duration? That crossed my mind and I experimented with 500ms, 750ms and 1 second (I ended up going with 750ms). I was thinking that it would help emphasize the tap behavior when developers look at it in the docs, but I don't have a preference either way.

That SGTM!

shihaohong:
Actually, I tried it again and 500ms did seem a little better :) I also took the opportunity to add `UniqueKey`s to each iteration since I noticed some that the state of the buttons between each example was being retained when it shouldn't be.

PR #103, opened 2019-08-24
Fixing default BinaryMessenger Breaking Change 

Piinks:
Fixing the diagram generator that was broken after changes introduced here:
https://github.com/flutter/flutter/pull/37489
https://github.com/flutter/flutter/pull/38464

This change ensures that the DiagramFlutterBinding is initialized.

### DIFF
--- a/packages/diagram_capture/lib/diagram_capture.dart
+++ b/packages/diagram_capture/lib/diagram_capture.dart
@@ -103,7 +103,7 @@ class _DiagramWidgetController extends WidgetController implements TickerProvide
   _DiagramWidgetController(WidgetsBinding binding) : super(binding);
 
   @override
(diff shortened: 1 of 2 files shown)


packages/diagram_capture/lib/diagram_capture.dart
```diff
@@ -146,7 +146,7 @@ class _DiagramTicker extends Ticker {
 
 // Provides a binding different from the regular Flutter binding so that
 // diagrams can control their timeline and physical device size.
```
gspencergoog: If you're going to make it "public", then you should probably change this to be a dartdoc comment.

gspencergoog:
Thanks for fixing that!

PR #114, opened 2020-05-11
Added Images of Slider and RangeSlider components and fixed analyzer issues

JoseAlba:
Added images of slider and range slider for better documentation. Also added some miner code changes that fixed the Flutter analyzer issue. 

Related Issue:
flutter/flutter#53290
flutter/flutter#56921


packages/diagram_capture/lib/diagram_capture.dart
```diff
@@ -150,8 +150,8 @@ class DiagramFlutterBinding extends BindingBase
     with
         GestureBinding,
         SemanticsBinding,
-        ServicesBinding,
         SchedulerBinding,
+        ServicesBinding,
```
rami-a, line 154: This change doesn't seem necessary.

JoseAlba: I know this was something I wanted to mention in what I learned this week. But services binding is a mixin on BindingBase and SchedularBinding. So for some reason I needed to put the SchedularBinding mixin first before the ServicesBinding mixin. If you look at the other mixins below ServicesBinding those also extend multiple classes.

PR #117, opened 2020-07-27
Fix device specification so you can specify the device name or the serial number

gspencergoog:
This changes the `--device` option to be `--device-id`, to match what `flutter` calls it.

Also, added a function to look up the id of the attached device from the given device name, so that the user doesn't need to know to put in the serial number, and can give the friendly name that flutter takes.

(I need to serial number because that's what `adb` uses to find a device.)

### DIFF
--- a/bin/generate.dart
+++ b/bin/generate.dart
@@ -327,7 +327,7 @@ class DiagramGenerator {
 
   Future<void> generateDiagrams(List<String> categories, List<String> names) async {
     final DateTime startTime = DateTime.now();
-    if (!await _findIdForDeviceName(device)) {
+    if (!await _findIdForDeviceName()) {
       stderr.writeln('Unable to find device ID for device $device. Are you sure it is attached?');
       return;
     }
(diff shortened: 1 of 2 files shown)


bin/generate.dart
```diff
@@ -316,8 +318,20 @@ class DiagramGenerator {
...
+  String deviceId = '';
+
+  /// The targetPlatform from the `flutter devices` output of the device we're
+  /// targeting.
+  String deviceTargetPlatform = '';
+
   Future<void> generateDiagrams(List<String> categories, List<String> names) async {
     final DateTime startTime = DateTime.now();
+    if (!await _findIdForDeviceName(device)) {
+      stderr.writeln('Unable to find device ID for device $device. Are you sure it is attached?');
```
HansMuller, line 331: Should `$device` be `deviceId`?

gspencergoog: No, that's where I convert the device into a deviceId.  I suppose I don't actually have to pass the device, since it's a member var.

bin/generate.dart
```diff
@@ -418,12 +454,13 @@ class DiagramGenerator {
...
-          // Yes, specify the -framerate flag twice: once for input, once for output.
+          // Yes, specify the -framerate flag twice: once for input, once for
+          // output.
           '-framerate', metadata.frameRate.toStringAsFixed(2),
           '-tune', 'animation', // Optimize the encoder for cell animation.
           '-preset', 'veryslow', // Use the slowest (best quality) compression preset.
-          // Almost lossless quality (can't use lossless '0' because Safari doesn't
-          // support it)
+          // Almost lossless quality (can't use lossless '0' because Safari
+          // doesn't support it)
```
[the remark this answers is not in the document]
gspencergoog: Fixed.

PR #119, opened 2020-07-29
This updates the generate script to use the `process_runner` package.

gspencergoog:
Now that I've published the `ProcessRunner` code as a separate package, this updates the generate script to use the package instead of its own copy.


Piinks:
I'm not sure though why the cla won't resolve. :(

PR #121, opened 2020-08-09
Updated to latest framework changes

darrenaustin:
Small updates to get things running properly on macOS:

- Update the `_DiagramWidgetController` to add a now needed `pumpAndSettle` method
- Updated the main in example to use `DiagramFlutterBinding.ensureInitialized()` so that it run properly
- Updated some build artifacts when running from macOS.

### DIFF
--- a/packages/diagram_capture/lib/diagram_capture.dart
+++ b/packages/diagram_capture/lib/diagram_capture.dart
@@ -131,6 +131,35 @@ class _DiagramWidgetController extends WidgetController implements TickerProvide
   Future<List<Duration>> handlePointerEventRecord(List<PointerEventRecord> records) async {
     return const <Duration>[];
   }
+
+  @override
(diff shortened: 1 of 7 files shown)


packages/diagram_capture/lib/diagram_capture.dart
```diff
@@ -131,6 +131,35 @@ class _DiagramWidgetController extends WidgetController implements TickerProvide
   Future<List<Duration>> handlePointerEventRecord(List<PointerEventRecord> records) async {
     return const <Duration>[];
   }
+
+  @override
+  Future<int> pumpAndSettle([
```
darrenaustin, line 136: @gspencergoog Not sure this if this is the best way to do this, but I cribbed it from the framework implementation.

gspencergoog:
I had to bash the CLA bot over the head, but all the checks pass now.

PR #122, opened 2020-08-10
Have Cirrus use master instead of dev for precommit checks

gspencergoog:
This moves Cirrus over to use the master branch as the one to sync for doing the CI checks, which avoids problems when trying to add diagrams that use a new Flutter API. It does mean that a broken master branch can also break a PR here, but hopefully that is fairly rare (and this repo is pretty low traffic, so not a huge cost in any case).


gspencergoog:
Actually, can you just add this to https://github.com/flutter/assets-for-api-docs/pull/121, since this is failing without those changes anyhow?

PR #127, opened 2020-12-07
Convert to null safety, update application shims

gspencergoog:
This converts the diagram generator to null safety, updates the application shims to the latest version, and updates the generated images to images generated with the new code.

Not all dependent packages are null safe yet (args, path_provider, etc), so I had to add `--no-sound-null-safety` in a few places.

I swapped out the version of `FakeProcessManager` for the one in the null safe version of `process_runner` and got rid of the dependency on Mockito.

Re-running the script to regenerate the images caused the floating_action_button_location images to change, but that is because a bug was fixed without re-running it. The new images look much better.


bin/generate.dart
```diff
@@ -86,23 +94,28 @@ class DiagramGenerator {
   /// targeting.
   String deviceTargetPlatform = '';
 
-  Future<void> generateDiagrams(List<String> categories, List<String> names) async {
+  Future<void> generateDiagrams(
```
darrenaustin, line 97: Were these added line break formatting changes intended, or an auto formatting thing? It looks like an auto formatter might have hit the whole file. If intended, I don't think they make these lines more readable, but YMMV 😏 .

bin/test/fake_process_manager.dart
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018 The Chromium Authors. All rights reserved.
+// Copyright 2020 The Flutter Authors. All rights reserved.
```
darrenaustin, line 1: Are we changing all the headers to Flutter copyright, or just this file? And 2020?

bin/test/fake_process_manager.dart
```diff
@@ -190,4 +242,4 @@ class StringStreamConsumer implements StreamConsumer<List<int>> {
     subscriptions.clear();
     return Future<dynamic>.value(null);
   }
-}
+}
```
darrenaustin, line 245: Nit: needs new line at end.

darrenaustin (approved):
LGTM. There is a lot here and I mostly glossed over the new platform files. All the other NNBD changes look great. I just had a couple of formatting nits, but nice job.

darrenaustin:
Oh and obviously there are merge conflicts and the analyzer appears to be unhappy as well.

bin/generate.dart
gspencergoog: It was dartfmt.  Sorry, I must have hit the wrong shortcut in the editor.  I'll revert these.

bin/test/fake_process_manager.dart
gspencergoog: LOL, no.  I just copied this from another repo (process_runner).  Reverted copyright.

bin/test/fake_process_manager.dart
gspencergoog: Added.

PR #131, opened 2021-01-15
FontFeature diagrams

Hixie:
Adds a bunch of samples for FontFeature.


Hixie:
See https://github.com/flutter/engine/pull/23730 for what these document.

Hixie:
cc @GaryQian

utils/diagram_generator/lib/main.dart
```diff
@@ -62,6 +62,8 @@ Future<void> main() async {
   final List<String> categories = flags['category'] as List<String>;
   final List<String> names = flags['name'] as List<String>;
 
+  print('Filters:\n  categories: $categories\n  names: $names');
```
GaryQian, line 65: Nit: Was this meant to be for debug or actually included here?

utils/diagram_generator/pubspec.yaml
```diff
@@ -26,10 +26,50 @@ environment:
...
+      fonts:
+        - asset: fonts/SortsMillGoudy-Regular.ttf
+
+    - family: Source Code Pro
+      fonts:
+        - asset: fonts/SourceCodePro-Bold.ttf
+
+    - family: Ubuntu Mono
+      fonts:
+        - asset: fonts/UbuntuMono-Bold.ttf
```
GaryQian, line 75: Nit trailing newline

GaryQian (approved):
LGTM, looks really comprehensive!

utils/diagram_generator/lib/main.dart
Hixie: It was intended to stay. The generator is really only run by us, so it's useful to see what it's doing. Previously we didn't report what filters we applied. (This is especially helpful on Linux where we don't currently handle arguments so the filters you specify on the command line don't actually apply and previously that was very confusing.)

utils/diagram_generator/pubspec.yaml
Hixie: good catch, fixed

Hixie:
Thanks, will land on green.

PR #133, opened 2021-02-20
Update body text property name

marwenx:
Updates the name of the body text property from the deprecated `body1` to `bodyText2` in the asset `material_app_theme_data`. This matches the description in the [api docs page](https://api.flutter.dev/flutter/material/ThemeData-class.html#material.ThemeData.2) for `ThemeData`. ([source code](https://github.com/flutter/flutter/blob/9b2d32b605/packages/flutter/lib/src/material/theme_data.dart#L154))

| Before        | After
| :-------------: |:-------------:|
| material_app_theme_data (original)    | material_app_theme_data |


goderbauer (approved):
LGTM

HansMuller:
This example and its diagram are out of date. The AppBar and FAB's colors default to the theme's ColorScheme. The updated digram is an improvement, so LGTM. Both the example and the diagram do need another update though.

PR #134, opened 2021-02-23
add FilterQuality diagram

yjbanov:
Add a diagram showing the effect of `FilterQuality` on scaled images.

Bonus: exit with a non-zero code when any of the worker jobs fail.


bin/generate.dart
```diff
@@ -348,11 +349,33 @@ class DiagramGenerator {
...
     }
   }
 }
 
+/// Sets the exit code to 1 if at least one of the jobs failed.
+void _checkJobResults(List<WorkerJob> jobs) {
+  if (jobs.any(_hasJobFailed)) {
+    // We're already printing the error message. All that's left is set
+    // the exit code.
+    exitCode = 1;
```
Hixie, line 363: can you return 1 here and have the exitCode be set outside this? In general code is easier to reason about if it does not cause changes outside of its own state.

Hixie (approved):
LGTM modulo minor comments
LGTM

bin/generate.dart
yjbanov: Done. Consolidated error reporting into `GeneratorException`.

Hixie: that works. thanks.

PR #136, opened 2021-03-10
Leading distribution documentation assets

LongCatIsLooong:
https://github.com/flutter/flutter/issues/72521

Breakdown of vertical components `line-height` comprises, and side-by-side comparison of common text height configurations. I'll explain what "Configuration X" is in the API docs:
- Configuration 1: `TextLeadingDistribution.proportional`
- Configuration 2: `TextLeadingDistribution.proportional` and `applyHeightToFirstAscent = false`
- Configuration 3: `TextLeadingDistribution.even` 
- Configuration 4: `TextLeadingDistribution.even` and `applyHeightToLastDescent = false`

The captions are slightly illegible at 30% scale:


packages/diagrams/lib/src/text.dart
```diff
@@ -484,6 +486,514 @@ class TextHeightComparisonPainter extends CustomPainter {
...
+        color: Colors.black,
+      ),
+    );
+    labelPainter.layout();
+    labelPaintOrigin = Offset(0, baseline) + fontMetricsLabelPadding;
+    labelPainter.paint(canvas, labelPaintOrigin);
+
+    // Top
+    labelPainter.text = TextSpan(
+      text: 'Line-Box Top',
```
LongCatIsLooong, line 974: Not sure this is the best term. "line box" in css seems to have a different meaning:  https://www.w3.org/TR/CSS2/visuren.html#inline-formatting

> The rectangular area that contains the boxes that form a line is called a line box.

packages/diagrams/lib/src/text.dart
```diff
@@ -484,6 +486,514 @@ class TextHeightComparisonPainter extends CustomPainter {
...
+  late final TextPainter heightCaptionTextPainter = TextPainter(textDirection: textDirection);
+  static late final TextStyle heightCaptionTextStyle = TextStyle(fontSize: 20, color: Colors.blue[900]!);
+  static const Offset fontMetricsLabelPadding = Offset(0, 3);
+  static const double heightCaptionBracketMinX = -25;
+
+  @override
+  Size computeDryLayout(BoxConstraints constraints) {
+    textPainter.layout(minWidth: constraints.minWidth, maxWidth: double.infinity);
+    if (shouldPaintHeightIndicator) {
+      heightCaptionTextPainter.text = TextSpan(text: 'line-height:\n${textPainter.height} px', style: heightCaptionTextStyle);
```
justinmc, line 749: The use of hyphens here for line-height and line-box makes me think that this is referring to something specific in Flutter (or it makes me think of CSS).  It might be confusing to users when they see this diagram show that line-height is 360px, but there is no direct line-height or lineHeight parameter on EditableText that can be used to set it to 360px.  In the docs that use this diagram, we should be clear about where these values come from and how they can be achieved in Flutter.

packages/diagrams/lib/src/text.dart
justinmc: Yeah it seems very similar but not quite.  As I understand it, the CSS definition would be the bounding height of all inline elements in the line, but as you've defined it, it's the local height of the current span.

Related to my comment above about hyphens, what would this be in Flutter?  Is it the same as the "intrinsic height" of the render box?

justinmc (approved):
I left some comments that might help, but I'm not sure.  Either way LGTM 👍 

My main thought is just that we need to make sure users can make the connection between this diagram and how to achieve what they want using the configuration available to them in Flutter.  I think we just need to be really clear about how to tweak the labeled parameters in an actual TextField, etc.  That is the responsibility of the docs where this diagram will be used, outside of this PR, though.

One question: Is it invalid to set TextLeadingDistribution.proportional and applyHeightToFirstAscent/applyHeightToLastDescent false?  Just seeing that those are missing from the given 4 configurations.

By the way, this diagram is really cool.  We need a lot more stuff like this in the docs!

packages/diagrams/lib/src/text.dart
LongCatIsLooong: Ah good point. How do "Text Height"  and "Text Top/Bottom" sound?

packages/diagrams/lib/src/text.dart
```diff
@@ -484,6 +486,514 @@ class TextHeightComparisonPainter extends CustomPainter {
...
+  Widget build(BuildContext context) {
+    return Container(
+      alignment: Alignment.center,
+      height: 500,
+      child: Column(
+        mainAxisAlignment: MainAxisAlignment.spaceBetween,
+        children: <Widget>[
+          const Center(
+            child: Text(
+              'Roboto, fontSize:$fontSize, height:$height',
```
goderbauer, line 508: nit: spaces after the `:`?

packages/diagrams/lib/src/text.dart
```diff
@@ -484,6 +486,514 @@ class TextHeightComparisonPainter extends CustomPainter {
...
+                  TextHeightBreakdownRow(
+                    text: text,
+                    backgroundColor: Colors.transparent,
+                    style: TextStyle(
+                      fontSize: fontSize,
+                      height: height,
+                      color: Colors.black,
+                      leadingDistribution: TextLeadingDistribution.proportional,
+                    ),
+                    shouldPaintHeightIndicator: true,
```
goderbauer, line 530: nit: remove the `should` from these params?

packages/diagrams/lib/src/text.dart
```diff
@@ -484,6 +486,514 @@ class TextHeightComparisonPainter extends CustomPainter {
...
+        _shouldPaintHeightIndicator = shouldPaintHeightIndicator,
+        _shouldPaintLeadingIndicator = shouldPaintLeadingIndicator,
+        _shouldPaintCaptions = shouldPaintCaptions,
+        textPainter = TextPainter(
+          textDirection: textDirection,
+          text: text,
+          textHeightBehavior: textHeightBehavior,
+        );
+
+  final TextPainter textPainter;
```
goderbauer, line 675: Can this be private?

LongCatIsLooong:
@justinmc @goderbauer addressed the review comments. also:
- `Line-Box Top/Bottom` => `Text Top/Bottom`
- `line-height` => `Text Height`
- Change the configurations used for demo so it looks more balanced

justinmc (approved):
Changes LGTM.

I like the new naming better.  It doesn't make me think that the labels are corresponding to something specific in Flutter/CSS this way.

PR #138, opened 2021-04-09
Clean up last of null safety migration, and CI setup.

gspencergoog:
This PR is a cleanup PR for the repo.  It does the following:

- Removes the last vestiges of the null safety migration so that everything runs with sound null safety.
- Moves CI scripts into the "ci" subdir, and makes some of them more general (so that they pick up new directories to analyze automatically, for one thing).
- Updates the application shims for the example app and the diagram generator app.
- Fixes a small issue in the generator where it wasn't using the passed process manager.


darrenaustin (approved):
LGTM. There is a lot here, so I just gave it a once over. Didn't see anything out of order for this kind of change.

PR #139, opened 2021-04-19
Update diagrams and fix some runtime errors caused by null safety migration

gspencergoog:
## Description

This updates all of the diagrams run with the latest framework, and fixes some runtime errors caused by nullsafety migration.

I've verified that the diagrams and videos look sane, but spot checks wouldn't hurt.

There are also some small code changes which need reviewing.

One thing I've noticed in doing this is that we might need to change some classes in the framework to handle nullable values (e.g. `AlignTransition`, and `RelativePositionedTransition`, and maybe others), because changing the type of the animation to use nullable values wasn't sufficient, and I had to create my own Tweens that were non-nullable to make it work.


Piinks (approved):
LGTM! There was a similar recent null safety change to BorderRadiusTween (https://github.com/flutter/flutter/pull/79860), perhaps we can audit.

PR #145, opened 2021-05-05
Sort imports automatically when creating snippets

gspencergoog:
## Description

This PR adds a utility for sorting imports according to the rules that the analysis server uses. It repurposes analysis server code so that it doesn't need the analysis_server package (which currently requires sdk < 2.0.0).

This way we won't need to suppress the `directives_ordering` lint when doing snippet analysis, and the snippets will have proper import ordering, even if the template and the sample source both have imports.

It pulls comments associated with the import along with them, to keep the comments in proper order as well.

Will help with https://github.com/flutter/flutter/issues/81697 once snippets are generated using this package.

## Related Issues

- https://github.com/flutter/flutter/issues/86197


pq:
In an ideal world you wouldn't need to duplicate this and could either leverage the source directly or fire up an instance of the analysis server and send it import source to organize.  It may be worth investigating options.  Happy to kick off a conversation w/ analyzer folks if you want.

gspencergoog:
I agree that it would be ideal not to duplicate this: my preference would be to be able to instantiate an `ImportOrganizer` from the analyzer package and just send in an analysis unit to be organized.

The `ImportsVerifier' in the analyzer mentions this in the source as something that missing: https://github.com/dart-lang/sdk/blob/cc1f9b9564144ca5d40bbe21b88acfcd461cf56a/pkg/analyzer/lib/src/error/imports_verifier.dart#L195

goderbauer:
+1 to finding a solution that doesn't require this duplication.

gspencergoog:
@goderbauer So, would you rather I not introduce this PR at all?  It may take some time before the analyzer has this capability: I'd rather duplicate it for now with a TODO, and remove it when the functionality is available.

goderbauer:
Does it generate any user-visible warnings (or confusion) when the imports are not sorted in the snippets?

goderbauer:
Maybe add some documentation to the code then indicating where this ic copied from and linking to some issue to fix the duplication?

gspencergoog:
@goderbauer PTAL, I've (finally) added that comment. Sorry, this one fell off my radar for a bit.

goderbauer (approved):
LGTM

PR #147, opened 2021-05-25
Pin analyzer version to avoid bad 1.6.0

gspencergoog:
This pins the analyzer to not allow 1.6.0, since the analyzer has a bad version that won't compile.

For reference: https://github.com/dart-lang/sdk/issues/46136


darrenaustin (approved):
LGTM. Thanks for the quick fix.

PR #150, opened 2021-06-26
Update more diagrams, fix bugs

gspencergoog:
## Description

Update the diagrams to latest. Standardizes the monospace font used for the blend diagrams to be "Courier New", as that is likely to be available on all platforms, and gets around the bug in https://github.com/flutter/flutter/issues/85360

Also turns off the debug banner on tabs video.

## Related Issues

- https://github.com/flutter/flutter/issues/85360


shihaohong (approved):
LGTM!

PR #151, opened 2021-07-15
Enable extracting of samples into separate files.

gspencergoog:
## Description

This modifies the snippets library to include functionality that can be used to extract the snippets into separate files and replace the code in the comments with a pointer to the new location.

This code is meant to be run once, really, to convert the repo to external comments, but needs to be runnable repeatedly so that I can keep from having merge conflicts with new code when I commit the change that switches things over.

This is one step in the process of converting to external snippets: the next will be to update this library to be able to publish docs using the source link instead of embedded source.


packages/animation_metadata/pubspec.yaml
```diff
@@ -1,6 +1,5 @@
 name: animation_metadata
 description: A package that just defines how metadata for an animation is held.
-author: Chromium Authors
```
darrenaustin, line 3: Is this just getting rid of a vestigial tie to the original source?

packages/snippets/bin/extract.dart
```diff
@@ -0,0 +1,142 @@
...
+      if (sample.type != 'dartpad' && sample.type != 'sample') {
+        continue;
+      }
+      final String relativePath = path.relative(sample.start.file!.path,
+          from: flutterInformation.getFlutterRoot().absolute.path);
+      snippetGenerator.generateCode(
+        sample,
+        includeAssumptions: false,
+        copyright: _kCopyrightNotice,
+        description: 'See description in the comments in the file:\n  $relativePath',
```
darrenaustin, line 106: Would the line number be helpful in the description here? There may be multiple samples in a file, right?

packages/snippets/bin/extract.dart
```diff
@@ -0,0 +1,142 @@
...
+            element.name.snakeCase,
+            sample.index.toString(),
+            'dart',
+          ].join('.'),
+        ]),
+      );
+      await outputFile.absolute.parent.create(recursive: true);
+      if (outputFile.existsSync()) {
+        print('File $outputFile already exists!');
+        exit(-1);
```
darrenaustin, line 124: Would using `errorExit` be better here, or is that just for arg parsing?

darrenaustin (approved):
LGTM. Just some minor questions.

packages/animation_metadata/pubspec.yaml
gspencergoog: Yes.  I removed it because of a new lint that says the author field is deprecated.

packages/snippets/bin/extract.dart
gspencergoog: No, I originally had that, but Michael pointed out that it wouldn't be a good idea because it'll get out of date as soon as the file changes.  Since the templates all have a comment that says "// Flutter code sample for {{element}}" in them, the actual result looks something like this:

```
// Flutter code sample for AboutListTile

// See description in the comments in the file:
//   packages/flutter/lib/src/material/about.dart
```

So they have a filename and a component name.  I could make it more explicit, with something like:
"See the description in the first sample listed for component AboutListTile of file packages/flutter/lib/src/material/about.dart", but I'm not sure it adds much.

packages/snippets/bin/extract.dart
gspencergoog: Yeah, maybe it would.

packages/snippets/bin/extract.dart
darrenaustin: Ah that makes sense. What you have above sounds good.

PR #152, opened 2021-07-15
Move Snippets main to function

gspencergoog:
This just encapsulates the main app for building snippets into a separate function so that it can be called from the Framework code, but maintained here.


gspencergoog:
Nevermind.  I can do this with pub global activate.

PR #154, opened 2021-07-28
Conform with command line for Flutter repo snippets tool, update analyzer in preparation for publish

gspencergoog:
This incorporates some changes made to the Flutter repo snippets tool, as well, as making the command line interface conform to what that tool uses. The goal is to make this feature complete for switching over the docs generation to use this library instead of the one checked in to the flutter repo.

- Added `--output-directory` argument so as not to hard code the location in the Flutter repo.
- Incorporated retry semantics for git branch detection, brought over from the Flutter repo.
- Added awareness of a FLUTTER_VERSION environment variable, which contains the output of `flutter --version --machine` so that the snippets tool doesn't need to run the flutter command when run as part of Dartdoc.
- Upgraded dependencies to use the 2.0.0 analyzer.
- Bumped the version to 0.2.0


packages/snippets/bin/snippets.dart
```diff
@@ -16,23 +19,47 @@ const String _kHelpOption = 'help';
...
 const String _kPackageOption = 'package';
 const String _kTemplateOption = 'template';
 const String _kTypeOption = 'type';
 
 const LocalFileSystem filesystem = LocalFileSystem();
 
-String getChannelName() {
+class GitStatusFailed implements Exception {
+  GitStatusFailed(this.gitResult);
+  final ProcessResult gitResult;
```
goderbauer, line 31: nit: maybe add some blank lines before and after this line for seperation and to improve readability?

packages/snippets/bin/snippets.dart
```diff
@@ -16,23 +19,47 @@ const String _kHelpOption = 'help';
...
-    throw 'git status exit with non-zero exit code: ${gitResult.exitCode}';
-  final RegExpMatch? gitBranchMatch = gitBranchRegexp
-      .firstMatch((gitResult.stdout as String).trim().split('\n').first);
-  return gitBranchMatch == null
-      ? '<unknown>'
-      : gitBranchMatch.namedGroup('branch')!.split('...').first;
+  final ProcessResult gitResult = processManager.runSync(<String>['git', 'status', '-b', '--porcelain'],
+      environment: <String, String>{
+        'GIT_TRACE': '2',
+        'GIT_TRACE_SETUP': '2'
```
goderbauer, line 54: Was this just for debugging? Or is this useful to keep?

packages/snippets/bin/snippets.dart
```diff
@@ -41,10 +68,27 @@ const List<String> sampleTypes = <String>[
...
+    }
+  }
+  return getChannelName();
+}
+
 /// Generates snippet dartdoc output for a given input, and creates any sample
 /// applications needed by the snippet.
 void main(List<String> argList) {
-  final Map<String, String> environment = Platform.environment;
+  stderr.writeln('Invoked as: snippets ${argList.join(' ')}');
```
goderbauer, line 89: Was this just for debugging? Or is this useful for production?

packages/snippets/bin/snippets.dart
```diff
@@ -145,30 +198,38 @@ void main(List<String> argList) {
...
   final String serial = args[_kSerialOption] as String? ?? '';
-  final List<String> id = <String>[];
+  late String id;
   File? output;
+  final Directory outputDirectory = filesystem.directory(args[_kOutputDirectoryOption] as String).absolute;
   if (args[_kOutputOption] != null) {
-    id.add(path.basename(
-        path.basenameWithoutExtension(args[_kOutputOption] as String)));
-    output = filesystem.file(path.absolute(args[_kOutputOption] as String));
+    id = path.basenameWithoutExtension(args[_kOutputOption] as String);
```
goderbauer, line 205: Side-note: We should probably enable the `cast_nullable_to_non_nullable` lint for this repository which would require an extra null-check here.

goderbauer (approved):
Tests for the newly added functionality?

Implementation LGTM.

packages/snippets/bin/snippets.dart
gspencergoog: It's useful to keep because we were having mysterious crashes in git on CI, and this will help debug that. Added a comment.

packages/snippets/bin/snippets.dart
gspencergoog: Oh! Yes, that should be removed.

packages/snippets/bin/snippets.dart
gspencergoog: OK, will do. I'll do that in a follow up PR to synchronize with the main repo analysis options.

packages/snippets/bin/snippets.dart
gspencergoog: OK, done.

gspencergoog:
Added test for the `FlutterInformation` class.

packages/snippets/bin/snippets.dart
gspencergoog: Actually, that lint is already enabled, and I don't get any warning here, which is probably because the `operator[]` for `ArgResults` returns `dynamic`, when it probably should return `Object`. My guess is that was too breaking for `args` to do.

goderbauer: Interesting. I wonder if the lint should also fire if you cast dynamic to something non-nullable...

goderbauer: Filed https://github.com/dart-lang/linter/issues/2828.

PR #155, opened 2021-07-28
Fix missing dev dependency and add some missing "!"s

gspencergoog:
Adding a missing dev dependency before publishing, and some `!` operators that probably should be there (but don't produce warnings because `ArgResults.operator[]` returns `dynamic`).


goderbauer (approved):
LGTM

PR #156, opened 2021-07-30
Make snippets conform to the old snippets tool by formatting output

gspencergoog:
## Description

This makes the snippets tool output conform more closely with the output from the original one in the Flutter repo.  This PR changes the output so that it is formatter with the Dart formatter, and so that it only outputs samples if they are `dartpad` or `sample` samples, not just snippets.

## Tests
 - Updated output tests to take into account the new format.


goderbauer (approved):
LGTM

PR #159, opened 2021-08-11
Fix snippet ids so that they don't have invalid characters in them, and use LineInfo

gspencergoog:
## Description

This PR does two things:
1) It fixes a problem where the IDs for snippets have "dart:ui" in them, which is an invalid character when the `flutter create` tool tries to download samples from the website.
2) Removes the interval tree that I was using to find line numbers and replaces it with `LineInfo.fromContent` from the analyzer package. Functionally they should be the same.

## Tests
- Added tests for the snippets tool command line.


packages/snippets/bin/snippets.dart
```diff
@@ -224,10 +224,10 @@ void main(List<String> argList) {
   } else {
     final List<String> idParts = <String>[];
     if (packageName.isNotEmpty && packageName != 'flutter') {
-      idParts.add(packageName);
+      idParts.add(packageName.replaceAll(RegExp(r'\W'), '_').toLowerCase());
```
goderbauer, line 227: > No need to add tests, since the functionality should be the same and existing tests cover it.

This does look like new functionality, though, that should probably be covered by a test?

goderbauer (approved):
Nice code deletion!

LGTM modulo the test.

packages/snippets/bin/snippets.dart
gspencergoog: Oh yeah, good point.  OK, one test coming up!

gspencergoog:
OK, I added a test for the main, and parsing of the command line options, which the thing needed anyhow.

PR #160, opened 2021-08-13
Convert sampler to be runnable with pub global run

gspencergoog:
## Description

This converts the sampler tool so that it can be run with `pub global run sampler` (after running `pub global activate sampler`, of course).

This is normally not possible for a Flutter app, but this gets around that restriction by having a command line application that runs `flutter build` to build the Flutter executable if it doesn't exist.

This only works on Linux, macOS, and Windows, since those are the only platforms that the sampler tool runs on.

(the diffs are a little misleading: I moved the previous `sampler.dart` into the `lib` directory as `main.dart`, and replaced the one in `bin` with the command line app).

This allows for easy distribution and running of the sampler Flutter utility.

I also rename the `extract` utility in the snippets package to `extract_sample` to make it a little less generic.


goderbauer (approved):
LGTM

Didn't review main.dart assuming that it is jus a copy of what was already reviewed.

gspencergoog:
Yes, that's correct: no changes to the main.dart.

PR #163, opened 2021-08-13
Fix test CI script to run ALL the tests, instead of just the ones it likes.

gspencergoog:
## Description

The `testing.sh` CI script was a little deficient, and was only running a few of the tests, not all of them.

Now it will run any tests it sees, as well as testing the published packages for publishability.


goderbauer (approved):
LGTM

PR #164, opened 2021-08-16
Move sampler compilation into a temporary directory.

gspencergoog:
## Description

On my previous change (#160) I was only able to test it locally via "pub global activate --source path sampler", which is using the source in place rather than installing it in the pub cache. Turns out that writing the build the pub cache did some weird things to pub, and Flutter refuses to run `pub get` inside of one (sensibly).

So, instead of doing the build in place, I'm now copying the code to a temp dir and building it there. Yes, this means that you'll have to recompile if you reboot, but that's not a huge price to pay, and you'll have to recompile if you update anyhow.

I'm not super happy with the performance of the startup the first time (it takes about a minute), but subsequent times are reasonable.

I also eliminated the platform directories and generate them in the temp dir with the `flutter` used to do the building.


utils/sampler/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Change Log for `sampler`
 
+## 0.3.0
+
+* Run the globally activated executable out of a temp dir instead of the 
+  pub cache
+* Remove most of the platform directories (leave macOS entitlements), and 
+  run `flutter create` on the platform to create them.
```
goderbauer, line 8: Do you have tests in place that detect when changes to the template (and/or the flutter create behavior) break the app?

(That has been a source of undetected breakages for things in flutter/test where a similar approach was used)

utils/sampler/bin/sampler.dart
```diff
@@ -78,38 +80,112 @@ String getPackageDigest(Directory packageLocation) {
...
+      Directory newDirectory = destination.absolute.childDirectory(entity.basename);
+      await newDirectory.create();
+      await copyDirectory(entity.absolute, newDirectory);
+    } else if (entity is File) {
+      await entity.copy(destination.childFile(entity.basename).path);
+    }
+  }
+}
+
+void copyDirectorySync(Directory source, Directory destination) {
```
goderbauer, line 95: This method uses synchronous file API (listSync, createSync, copySync), but the method above used the asynchronous versions. Why this difference?

goderbauer (approved):
LGTM

utils/sampler/bin/sampler.dart
gspencergoog: Oh, actually, that was just because when I was debugging, the async version was just exiting (turns out I wasn't awaiting `main`), so I made a sync version.  I'll revert to the async version.

utils/sampler/CHANGELOG.md
gspencergoog: No, because I can't think of a good way to test it. I mean, I could try just running a test app in a temp dir with the same entitlements, I guess.

PR #166, opened 2021-08-20
Update CI scripts, fix Dockerfile to add flutter build prerequisites

gspencergoog:
## Description

This cleans up some of the CI scripts so that they're more useful to developers working on the repo, since they're not directly being run anymore as part of the CI.

Changes in this PR:
- Removed `ci/analyze.sh`, since it's now rolled into `check`
- Added `ci/fix_format` to fix any formatting issues in the repo in one command.
- Replaced `testing.sh` with `check`, which uses the same `flutter_plugin_tools` that the CI setup now uses, allowing the developer to specify any of `license-check`, `analyze`, `pubspec-check`, `version-check`, `publish-check`, or `test` (or run them all if none are specified).
- Renamed `pub_upgrade.sh` to `pub_upgrade` to keep in line with the other scripts.
- Renamed `tool_runner.sh` to `tool_runner` to keep in line with the other scripts.
- Added a README.md for the ci directory.


Piinks (approved):
NICE - LGTM 👏 👏 👏

.ci/Dockerfile
```diff
@@ -10,9 +10,9 @@ RUN sudo apt-get update && \
 RUN wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add - && \
     sudo apt-add-repository "deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-12 main" && \
     sudo apt-get update && \
-    sudo apt-get install --yes --allow-unauthenticated libffi7 && \
-    sudo apt-get install --yes --allow-unauthenticated libllvm12 && \
-    sudo apt-get install --yes --allow-unauthenticated libclang-cpp12 && \
-    sudo apt-get install --yes --allow-unauthenticated clang-format-12 && \
+    sudo apt-get install --yes --allow-unauthenticated libffi7 libllvm12 libclang-cpp12 clang clang-format-12
+
+RUN sudo apt-get install --yes cmake ninja-build pkg-config libgtk-3-dev && \
```
dkwingsmt, line 15: Why do we split this into a separate RUN?

gspencergoog: It doesn't need to be, but I separated it because the ones in the previous run all have to do with LLVM, and getting them from the LLVM repo. I figured it makes sense to group them because if they don't change, we don't need to rebuild that chunk of the image, and the rebuild of the Docker image can go faster, since it'll be cached.

PR #167, opened 2021-08-20
Remove extraneous files

gspencergoog:
This just removes some deleted files that snuck back in when I was doing some merges.


Piinks (approved):
Sweet! LGTM!

PR #168, opened 2021-08-27
Publish the 'extract_sample' command

gspencergoog:
## Description

Somehow the extract_sample command never got published in the pubspec.

This bumps the version, and publishes the command so that it can be run with `dart pub global run snippets:extract_sample`.

## Issues

- Addresses part of https://github.com/flutter/flutter/issues/89068

### DIFF
--- a/packages/snippets/CHANGELOG.md
+++ b/packages/snippets/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.2.4
+
+* publish the extract_sample command.
+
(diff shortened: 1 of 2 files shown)


creativecreatorormaybenot:
@gspencergoog Now, the command is `snippets:extract_sample` and still requires the `--input` flag, so the README still does not indicate it correctly, I updated it in https://github.com/flutter/flutter/pull/89065 👍

PR #170, opened 2021-10-01
Clean up and make --template optional

gspencergoog:
## Description

This makes the `--template` argument optional for the snippet tool, so that samples that have linked source files don't have to supply it, since it isn't needed.

Also, cleaned up some loose ends in the sample extractor, added support for parsing mixins (which was apparently missing), updated the GCP credentials for the CI, and fixed usages of `dart pub`.

### DIFF
--- a/packages/snippets/bin/extract_sample.dart
+++ b/packages/snippets/bin/extract_sample.dart
@@ -45,13 +45,13 @@ Future<void> main(List<String> argList) async {
     _kHelpOption,
     defaultsTo: false,
     negatable: false,
-    help: 'Prints help documentation for this command',
+    help: 'Prints help documentation for this command.',
   );
   parser.addFlag(
     _kVerboseOption,
     defaultsTo: false,
     negatable: false,
-    help: 'Prints extra output diagnostics',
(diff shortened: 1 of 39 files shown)


gspencergoog:
The GCP credentials are bad: looking into replacing them.

gspencergoog:
Replaced the credentials.

packages/snippets/bin/extract_sample.dart
```diff
@@ -46,21 +47,29 @@ Future<void> main(List<String> argList) async {
     negatable: false,
     help: 'Prints help documentation for this command',
   );
+  parser.addFlag(
+    _kVerboseOption,
+    defaultsTo: false,
+    negatable: false,
+    help: 'Prints extra output diagnostics',
```
goderbauer, line 54: nit: Some of the help texts here end with `.` and some don't. We should probably unify?

goderbauer (approved):
LGTM

packages/snippets/bin/extract_sample.dart
gspencergoog: Good point. Fixed.

PR #176, opened 2022-01-21
Fix the bit rot in this repo

gspencergoog:
## Description

Fixes the bit rot that has occured in this repo, so that CI tests pass cleanly.


gspencergoog:
Tests are failing because the service account token needs to be updated, as in: https://github.com/flutter/assets-for-api-docs/pull/175

yusuf-goog (approved):

PR #193, opened 2022-08-12
Fix analyzer warnings

goderbauer:
Clean up some analyzer warnings and pin the `analyzer` to version 4.3.1 because newer versions a deprecating API we currently use heavily. They will require a larger migration for which I filed https://github.com/flutter/flutter/issues/109473.

https://github.com/flutter/flutter/issues/109469


goderbauer:
This unblocks https://github.com/flutter/assets-for-api-docs/pull/192.

yusuf-goog (approved):

PR #196, opened 2022-08-25
Consistent theming across platforms

PixelToast:
This PR installs a DiagramMaterialApp widget that enforces a consistent theme across all platforms.

Unfortunately it's not perfect, there are issues with text aliasing on mac (perhaps related to https://github.com/flutter/flutter/issues/67034).

Mac on the left, Android on the right:

idea64_YSinmshsbn

Before landing I have a bunch more diagrams to check for consistency and would like to try out a potential solution to the thicker text.


Hixie:
Is the reason for this so that we can built the diagrams on a platform other than Linux?

Hixie:
cc @gspencergoog

PixelToast:
Yeah the goal is to make diagrams generated on iOS or Mac look the same as they do on Linux, currently they inherit the system's ThemeData.platform which leads to diagrams that look different. This PR works pretty well with the exception that font rendering looks significantly different on MacOS from the doubled aliasing.

packages/diagram_capture/lib/theme.dart
```diff
@@ -0,0 +1,71 @@
...
+/// textScaleFactor of 1.0.
+final MediaQueryData kDefaultMediaQuery = MediaQueryData(
+  size: kDefaultDiagramViewportSize,
+  devicePixelRatio: DiagramFlutterBinding.instance.pixelRatio,
+);
+
+/// A [MaterialApp] with defaults that are appropriate for capturing diagrams.
+///
+/// By default we emulate the theme of an Android device in light mode, and
+/// force the Roboto font.
```
gspencergoog, line 24: The  API docs website defaults to the "Open Sans" font, so it might match better than Roboto. I believe it has a similar license.

Hixie:
@PixelToast says "off with it's head"!

Hixie:
(and that we really should move to generating these assets in CI rather than manually, which is valid)

PR #198, opened 2022-09-08
Add diagrams for a bunch more canvas stuff

PixelToast:
This PR adds diagrams for Canvas.clipRect, Canvas.clipRRect, Canvas.clipPath, Canvas.drawLine, Canvas.drawRect, Canvas.drawRRect, Canvas.drawOval, Canvas.drawCircle, Path.conicTo, Path.cubicTo, Path.quadraticBezierTo, Radius.circular, and Radius.elliptical


packages/diagrams/lib/src/basic_shapes.dart
```diff
@@ -0,0 +1,1050 @@
...
+    );
+
+    canvas.restore();
+  }
+
+  @override
+  bool shouldRepaint(LineDiagramPainter oldDelegate) => true;
+}
+
+class RectCtorDiagramPainter extends CustomPainter {
```
gspencergoog, line 208:
```suggestion
class RectConstructorDiagramPainter extends CustomPainter {
```

We try not to use abbreviations in names.

gspencergoog (commented):
These are really awesome!

Can you perhaps change them so that they are transparent PNGs?  One thing that would be nice is if we can have them work for both light and dark mode docs pages, since that's something that they will soon respect.

It might be possible to choose (probably not black or white) colors/transparencies that will be readable over either a black background or a white background. I realize that might be a tall order, but since we don't have a good way to swap the assets based on the dark mode setting yet, seems like the best solution for now.

gspencergoog:
And if it isn't straightforward to generate an RGBA PNG (which it might not be), then we can also leave that for another time.

PixelToast:
It's totally possible to make them transparent! The white background was added intentionally to not break dark backgrounds.

I'm not sure if there's a good theme that works for both though, here's what I have so far:

image

image

image

image

gspencergoog:
I think that works great! Ship it!

PixelToast:
@gspencergoog re-generated them with the new color palette! let me know if we any further tweaks

gspencergoog (approved):

Thanks so much!

PR #206, opened 2022-12-05
Update assets

gspencergoog:
## Description

This just updates all of the assets, to get rid of some corrupted assets.

I've manually inspected all the images and spot checked the videos.

## Related Issues
 - Fixes https://github.com/flutter/flutter/issues/116393


Hixie:
LGTM for the images

didn't check the videos, RSLGTM for them

BoxFit.fillHeight has a trailing wrapped "t" we should probably fix at some point

We really need a better solution for reviewing these changes. @Piinks what would it take to hook this repo into skia gold, say?

Hixie (approved):

Piinks:
> what would it take to hook this repo into skia gold, say? 

It would require a separate instance of Gold to be created, since each instance is repo-based. Then we would likely be able to copy and re-use most of the flutter_goldens package, with some changes to grab the images generated by the tool here rather than the images from `matchesGoldenFile` like in the framework. As for the videos, Gold cannot support them.

PR #213, opened 2023-04-03
Updates to enable bin/generate to run again

HansMuller:
Corrected two problems that prevented bin/generate from running at all:
- DiagramFlutterBinding.updateDiagram attached its root widget without providing a MediaQuery. In the past the MediaQuery was provided by MaterialApp but that's no longer true. Now: `wrapWithDefaultView(rootWidget)`.
- rawExtendedRgba128 was added to ui.ImageByteFormat but the switch in _byteFormatToString didn't include it and was no longer null safe.


goderbauer (approved):
LGTM, but we will have to wait until both of these changes have made it into stable or change the analyzer check to run against master.

PR #226, opened 2023-08-28
Floating cursor gif

justinmc:
For a docs improvement around the floating cursor feature, in: https://github.com/flutter/flutter/pull/133002

I wasn't sure if I needed to run any generation scripts or anything after adding the file, please let me know!

Also, is it OK that this is 1. a gif and 2. not generated from a script?  I wanted an animation of the native floating cursor feature.


gspencergoog:
Normally we'd want the image to be generated, but in this case it's a capture of a native feature, so I don't see the need. This seems fine to me.

gspencergoog (approved):

Hixie:
In 5 years when iOS looks completely different, will we want to use this same diagram? Or does it need updating every few months for iOS changes?

Hixie:
(If the goal is to show the difference between the floating cursor and the background cursor, it might be ideal to move the cursor around much more slowly in the GIF; the way it spins right now at the start makes it quite distracting if you're trying to see what the cursor is doing, and it never really stays still for long enough to fully understand it.)

ricardoamador:
Before anyone merges this can you please try adding the 'autosubmit' label to see if the bot can merge it?

godofredoc:
@justinmc we just rolled out autosubmit to this repo. Please add the autosubmit label to merge.

gspencergoog:
Justin is OOO until January.  @Hixie are you OK with this merging, or do you want to do something different?

Hixie:
In the absence of a reason to do otherwise, I would feel more comfortable if this was an easily reproducible automated recording that used a slower pace than in the current GIF (per my comments above).

justinmc:
I'm going to close this and explain the floating cursor in https://github.com/flutter/flutter/pull/133002 without a visual aid. It's not possible to generate this image automatically because it requires interacting with a specific part of the soft keyboard.

PR #234, opened 2023-12-21
Re-enable bin tests

gspencergoog:
## Description

Somehow the tests in the bin/ directory were not being run. I'm not sure if they were ever run automatically, but they will be now...


bin/generate.dart
```diff
@@ -45,7 +45,7 @@ class DiagramGenerator {
   /// The path to the dart program to be run for generating the diagram.
   static final String generatorDir = path.join(
     projectDir,
-    'packages',
```
chunhtai, line 48: shouldn't this follow the folder structure?

In my pr I updated the test instead

PR #238, opened 2024-01-31
Add grass/remove burlap

gspencergoog:
## Description

Adding a grass image and removing the burlap image since it aliases too much.


HansMuller (approved):
LGTM

PR #249, opened 2024-07-15
Move and convert Engine doc assets

jtmcdole:
Part of flutter/flutter#151691

- Convert png to avif (supported by Safari, Firefox, and Chrome)
- Convert gif to webp (supported by Safari, Firefox, and Chrome)

Tool used: imagemagick and zsh scripting.


nate-thegrate (approved):
I thoroughly analyzed every `1` and `0` in each binary file, and they all look great.

jtmcdole:
I spot checked the images with transparency and that's when I switched from ffmpeg to imagemagick. ffmpeg's libav1stv is busted, and AOM was taking forever.  Thanks!

PR #250, opened 2024-08-05
Update a pubspec and platform runners

gspencergoog:
## Description

This updates a pubspec, all of the platform runners in the two packages that have platform directories.

The diagrams are updated separately in https://github.com/flutter/assets-for-api-docs/pull/251 including the contents of this change.


packages/diagram_generator/android/app/build.gradle
```diff
@@ -1,71 +1,44 @@
-// Copyright 2013 The Flutter Authors. All rights reserved.
```
goderbauer, line 1: Here and elsewhere: Do we need to keep these license headers?

goderbauer (approved):
RSLGTM

packages/diagram_generator/android/app/build.gradle
gspencergoog: The license check doesn't fail, and they're all in the generated code. I added them back in already to the ones that failed the license check. I'm assuming no.

PR #254, opened 2024-11-15
Merge queue screenshots

nate-thegrate:
I am not using my time productively, but this was very fun :)

- [x] removed background
- [x] higher image resolution
- [x] smaller file size


Piinks (approved):
LGTM, the formatter is blocking this though. Was going to follow up on it this week

